PR #121, opened 2014-12-22
update to use minified jquery

NPrescott:
Made sense to me to use the minified version of jquery since Frog is already using the minified copy of Bootstrap. Relatively minor change (from ~276KB to ~93KB) but I think it is a more sensible default to provide.

### DIFF
--- a/example/_src/page-template.html
+++ b/example/_src/page-template.html
@@ -102,7 +102,7 @@
       </footer>
     </div>
     <!-- </body> JS -->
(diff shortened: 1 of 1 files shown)


greghendershott:
Sorry for the delay during the holidays to respond to this.

Good idea. On the one hand, the `example` directory is intended as only that, and people really ought to source the Bootstrap files directly themselves, as the Frog README explains. On the other hand `raco frog --init` does use some files from `example` as a quick-start, and as you say, it may as well use the minified version.  So: Merged. Thanks!

PR #125, opened 2015-05-13
Add config param for python executable

RenaissanceBug:
* Add default & explanation to the .frogrc
 * Add python-executable param to frog.rkt and params.rkt
 * Add explicit location of executable and an existence check/warning to pygments.rkt

I made these changes because I'd installed Pygments under python3 on my machine, but the default in my path was Python 2 and Frog consequently couldn't find Pygments.


greghendershott:
Thanks!

Issue #119, opened 2014-12-21
Make it easy to support blogs rooted other than at /


tfeb:
I have an existing site, and I want to use frog to create a blog hanging below a directory in that site, without perturbing the existing structure  So I want to create a structure like
- `<root>/index.html` -- main index
- `<root>/feeds/...` -- feeds

and so on: I want absolutely everything to hang beneath `<root>`.

It is at best painful to do this at present, and I'm not sure it's completely possible, because things like feed links seem to have hardwired URLS which look like `/feeds/...`.  It may be that this is just a documentation (or comprehension on my part) problem, but I'm not sure.  In any case it would be cool if this was easier to do.

greghendershott:
Thank you for working on this! I wanted to comment and let you know I haven't forgotten about it. I did read through your changes. They seem reasonable and I don't see any obvious problems or omissions. But -- my brain had a cache miss on some of the details of what Frog is doing. :) So, I'll need to find a chunk of time to reload, and to study it more carefully.

tfeb:
That's fine of course.  For what it's worth I've been using it for a while and it empirically works though this doesn't mean it's actually the right answer! I've been rebasing my branch on master and can push that if/when you want to do a pull (I know push/rebase/push is evil but it can make things easier).

Let me know if/when you want me to make a pull request.

greghendershott:
Done with #126 and additional commits.

PR #126, opened 2015-05-18
Support rerooting frog

tfeb:
Hi, I've made some changes to support rerooting frog (discussed in a previous issue I think), so blogs do not need to hang under `/`.  I'm currently maintaining my own fork with these changes: it would be a good thing, for me at least, if they were merged into the real version.  They are reasonably well tested, but I didn't understand any more of frog than I needed to in order to make them (I am reasonably familiar with Racket, so the code should not be too horrid), and haven't poked around enough since to understand if they fit into the architecture.

As such you might not want to merge *these* changes, but I think something *like* them would be good: hence this request, before I forget it.  Being slightly more pragmatic, these changes do work, don't break any existing functiionality, and are reasonably easy to drive (there's just a `uri-prefix` parameter which drives everything from the config file), so they might well do in the interim.

The two important commits are

- b3bf6e3 is the support itself.
- 322929f lets you reroot the preview server -- you want this if you ever use it with rerooted blogs.

There is a slightly spurious commit (6d562a2) whose job is just to deal with Emacs backups: it's not related to the others, although it's useful (to me) and I think innocuous. It should be on another branch really but I got lazy.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,12 @@ example/tags/
 example/2*
 example/A-Non-Post-Scribble-Page*
 example/img/posts/
+
+# Emacs / DrRacket backups & auto save files
+*~
+\#*\#
+.\#*
+
+# Emacs tags
+TAGS
+
--- a/example/.frogrc
+++ b/example/.frogrc
@@ -1,10 +1,10 @@
 # Required: Should NOT end in trailing slash.
 scheme/host = http://www.example.com
 
-# Prefix for URIs, default being /
-# this will be interposed at the beginning of any URI, including those
-# specified here.
-#uri-prefix = /
+# A path prepended to URIs, including those specified here in .frogrc
+# such as `permalink` and `posts-index-uri`. Defaults to `/`. This is
+# useful when you want to embed your blog in another web site.
+uri-prefix = /
 
 # The title of the blog. Used when generating feeds.
 title = My Awesome Blog
--- a/example/.frogrc
+++ b/example/.frogrc
@@ -1,6 +1,11 @@
 # Required: Should NOT end in trailing slash.
 scheme/host = http://www.example.com
 
+# Prefix for URIs, default being /
+# this will be interposed at the beginning of any URI, including those
+# specified here.
+#uri-prefix = /
+
 # The title of the blog. Used when generating feeds.
 title = My Awesome Blog
 
--- a/frog/bodies-page.rkt
+++ b/frog/bodies-page.rkt
@@ -71,11 +71,11 @@
                     ,@(if (current-show-tag-counts?) `(,(format "(~a)" v)) '())
                     " "
                     (a ([href ,(atom-feed-uri k)])
-                       (img ([src "/img/feed.png"]))))))
+                       (img ([src ,(canonicalize-uri "/img/feed.png")]))))))
     (p (a ([href ,(current-posts-index-uri)]) "All Posts")
        " "
        (a ([href ,(atom-feed-uri "all")])
-          (img ([src "/img/feed.png"])))))))
+          (img ([src ,(canonicalize-uri "/img/feed.png")])))))))
 
 (define (tags-list-items)
   (for/list ([(k v) (in-dict (tags-alist))])
@@ -95,7 +95,7 @@
                         ", ")))
 
 (define (tag->xexpr s)
-  `(a ([href ,(str "/tags/" (our-encode s) ".html")]) ,s))
+  `(a ([href ,(canonicalize-uri (str "/tags/" (our-encode s) ".html"))]) ,s))
 
 (define (blurb->description s)
   (~> (with-input-from-string s read-html-as-xexprs)
--- a/frog/feeds.rkt
+++ b/frog/feeds.rkt
@@ -27,7 +27,7 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
 (define (atom-feed-uri tag)
-  (str "/feeds/" tag ".atom.xml"))
+  (canonicalize-uri (str "/feeds/" tag ".atom.xml")))
 
 (define (write-atom-feed xs title tag of-uri-path file)
   (prn1 "Generating ~a" (abs->rel/www file))
@@ -41,7 +41,7 @@
       [xml:lang "en"])
      (title ([type "text"]) ,(str (current-title) ": " title))
      (link ([rel "self"]
-            [href ,(full-uri (abs->rel/www file))]))
+            [href ,(full-uri (canonicalize-uri (abs->rel/www file)))]))
      (link ([href ,(full-uri of-uri-path)]))
      (id () ,(str "urn:"
                  (our-encode (current-scheme/host))
@@ -86,7 +86,7 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
 (define (rss-feed-uri tag)
-  (str "/feeds/" tag ".rss.xml"))
+  (canonicalize-uri (str "/feeds/" tag ".rss.xml")))
 
 (define (write-rss-feed xs title tag of-uri-path file)
   (prn1 "Generating ~a" (abs->rel/www file))
@@ -115,7 +115,8 @@
 
 (define (post->rss-feed-entry-xexpr tag x)
   (match-define (post title src-path modtime dest-path uri-path date older newer tags blurb more? body) x)
-  (define item-uri (full-uri/decorated uri-path #:source tag #:medium "RSS"))
+  (define item-uri (full-uri/decorated uri-path
+                                       #:source tag #:medium "RSS"))
   `(item
     ()
     (title ,title)
--- a/frog/frog.rkt
+++ b/frog/frog.rkt
@@ -36,6 +36,7 @@
   (parameterize* ([top (find-frog-root)])
     (parameterize-from-config (build-path (top) ".frogrc")
                               ([scheme/host "http://www.example.com"]
+                               [uri-prefix #f]
                                [title "Untitled Site"]
                                [author "The Unknown Author"]
                                [editor "$EDITOR"]
--- a/frog/frog.rkt
+++ b/frog/frog.rkt
@@ -108,8 +108,9 @@
         (set! port (string->number number))]
        [("--root") path
         (""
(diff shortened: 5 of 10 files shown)


willghatch:
Oh, geeze, I should have looked at the existing pull requests before adding this feature myself and adding another pull request for it.  But that said, clearly it shows that it's an important feature if two people wrote it.

tfeb:
willghatch: If I get a chance I'll look at your fork and see how it relates to mine, then may be we could organize a single combined pull request.  I agree this is something it needs, and I'm not particularly committed to mine other than that I might have to change config in my toy blog to use some other one.

willghatch:
@tfeb, frankly, yours is better.  Mine is more hackey in that the paths that are set up as strings I just add in a variable that may be a `.`, therefore giving stuff like /./foo.html.  I was thinking I would go back and add a function to clean it up, like your `canonicalize-url` function, but then I saw your PR.  Also I hadn't yet gotten to making the web server start with the right page up, which it looks like yours does (granted, I haven't actually tested yours yet).

greghendershott:
@tfeb and @willghatch Thank you very much for the pull requests! I apologize for not reviewing @tfeb 's yet. I've been working on other projects lately.

---

The original use case for Frog was a simple blog-only site. Beyond that, if people wanted to add a blog to an existing domain example.com, I imagined they'd use a subdomain like blog.example.com (in which case it's fine if everything hangs off `/` there).

Later I got a couple requests to add configuration enabling you to put the blog posts and blog post index page elsewhere:

```sh
# Pattern for blog post permalinks
# Optional: Default is "/{year}/{month}/{title}.html".
# Here's an example of the Jekyll "pretty" style:
permalink = /blog/{year}/{month}/{day}/{title}/index.html
```

```sh
# The URI for the index of blog posts. Defaults to /index.html but you
# could change to e.g. /blog/index.html or /posts-index.html, and use
# some other /index.html for your site. This also effects the URI used
# in the Atom and RSS feed files for posts.
posts-index-uri = /index.html
```

However that leaves some other things still hanging at root: images, feeds, tags (more?).

So that's the status quo, I think.

---

As for this pull request: At a quick glance it looks fine, and it's great you've used it for real and done things like adjust the preview server. Thanks!

My gut tells me there are one or more bugs lurking here, re interaction with other options, that I'm just not seeing. However I think the answer to that is that I should beef up the testing.  Ideally I'd do that _before_ merging this PR. However realistically maybe I ought to merge the PR, first.

Please let me know if you have any thoughts/comments about the above?  And thanks again for the PR!

tfeb:
I think it would better to do the testing stuff first, as there probably are bugs, and there's no advantage to merge code we think is probably broken.  I'm happy to keep tracking stuff and checking/merging against my branch as/when.  I'm never quite sure how to get GitHub to tell me about commits to repos I have pull requests against, but I can work that out I'm sure.

greghendershott:
@tfeb OK. I've started to work on creating a test of the permutations of parameters that determine the location and/or URI path of various things. For each such permutation, it would check certain output files: Do the files exist where expected, and, inside the files are the generated URIs as expected?

Once I'm satisfied with that, for the existing parameters, I'll try merging your branch and adding the `uri-prefix` parameter to the test mix.

tfeb:
Cool.  If there are problems or things I can help with let me know.

willghatch:
Cool.  Thanks, Greg.

greghendershott:
I made some progress on the testing but got detoured onto another issue; plan to resume work on this tomorrow.

Also for reference here's a link to the original issue #119: https://github.com/greghendershott/frog/issues/119

greghendershott:
@tfeb I switched from working on the tests to trying out your PR branch.

All works well for me, except when I try to use the preview server. Can you give me a example of the values you use for:

- `output-dir`
- `uri-prefix`
- the `--root` flag?

Maybe I'm being dense but I can't find a combination that works for me with `raco frog -p`.

greghendershott:
OK I had a typo in the `--root` arg.

What does work for me:

Use the frog example project. In its `.frogrc`, specify `output-dir = baz` and `uri-prefix = /baz`. Then, from the `example` project dir (same as where the `.frogrc` is) I do `raco frog --root . -p`.

That gets me a generic Racket web server page, which makes sense as there is no index.html up in example. But when I nav to `localhost:3000/baz/` I get the index.html down there, and navigation via links works thereafter.

tfeb:
I think what you're doing is what should work.  Unfortunately I can't quite remember as I now have a bunch of `make` rules which do it all for me, and I've forgotten what I changed.  I will look at it either this evening or tomorrow (UK time).

greghendershott:
@tfeb I pulled your PR branch locally so I could work on it a little bit, before eventually merging to master.

I pushed a couple commits [here](https://github.com/greghendershott/frog/commits/tfeb-uri-root-fix). Although I think they're pretty straightforward, if you have time to look at them and comment please feel free. Otherwise no worries.

tfeb:
I think your commits are fine (I didn't know about guards on parameters!).  I am confused about branches though (almost certainly my fault).  I fetched from your repo into my local copy assuming that `tfeb-uri-root-fix` would be properly ahead of `uri-root-fix` (so once I'd fetched and made a local branch I could FF mine onto it) but it shows up as beside it (equivalent commits, different hashes).  So I've now cherry-picked *those* commits onto the original branch and pushed that.  Almost certainly this is my poor branch management: in any case it doesn't matter I think.

For the `--root` thing: what `--root` is meant to do is to tell the preview server where `/` is, so that links work.  If it doesn't know that, and the blog is rerooted, then any non-relative link breaks.  What it *should* do, in the `--preview` case is point the browser at the blog home page as well, rather than at the root.  It doesn't currently, and that's a bug.

The clean-breaking thing is clearly a bug I think: it's not clear that clean can *work* when you've changed the root, but it should not blow up.

greghendershott:
>  I am confused about branches though (almost certainly my fault). I fetched from your repo into my local copy assuming that tfeb-uri-root-fix would be properly ahead of uri-root-fix (so once I'd fetched and made a local branch I could FF mine onto it) but it shows up as beside it (equivalent commits, different hashes). So I've now cherry-picked those commits onto the original branch and pushed that. Almost certainly this is my poor branch management: in any case it doesn't matter I think as either branch will do.

Well I'm not a master of git, and I didn't explain what I was doing.

I had made a new branch, pulled from yours, and rebased it (the last step is why the commit IDs changed).

I did this because I figured I wouldn't use the GitHub merge button. Instead I figured I'd add a few more commits, merge to master (I prefer linear, fast-forward merges), push, and mark the PR as closed manually.

greghendershott:
> For the --root thing: what --root is meant to do is to tell the preview server where / is, so that links work. If it doesn't know that, and the blog is rerooted, then any non-relative link breaks.

That makes sense. I just got confused because I didn't start with a clear picture of the correct result. When I did get it right, I thought I'd done it wrong when preview opened on the non-existent true root page instead of the blog root page.

>  What it should do, in the --preview case is point the browser at the blog home page as well, rather than at the root. It doesn't currently, and that's a bug.

I'll add that.

greghendershott:
I pushed commit 45b9a18.

For this to work, you still need to supply the `--root` command line flag. I wish that flag weren't necessary at all, and it could do the right thing automatically. Mulling that....

greghendershott:
What do you think about the following code to pick a default value of the `--root` flag:

```
      (define root
        ;; Default the server root to be the number of parent dirs
        ;; above (www-path) as there are dirs in current-uri-prefix.
        (let ([depth (sub1 (length (explode-path (current-uri-prefix))))])
          (simplify-path (apply build-path (list* (www-path)
                                                  (build-list depth (λ _ 'up)))))))
```

This works for the simple cases I tried:

- `uri-prefix = /`, `output-dir = .`, `www-path = /path/to/files/`:
    - The root defaults to `/path/to/files/`
- `uri-prefix = /baz`, `output-dir = baz`, `www-path = /path/to/files/baz/`:
    - The root defaults to `/path/to/files/`

What do you think?

---

On principle I don't like `..` or `'up` with web servers -- escaping up a directory structure makes me nervous.  But:

- Realistically, most people will be using `--preview`. I highly doubt anyone would use `--serve` in production. After all, Frog is billed as a _static_ blog generator not a live blog server.

- Anyway  the `--root` flag is still available to use; this proposed change simply picks a default value for it.

greghendershott:
Actually I may as well push that to this topic branch as a commit. See commit 805792a

greghendershott:
Derp that commit had two problems. Need to handle `(current-uri-prefix)` being `#f`. Also, `explode-path` is provided by `racket/base` only starting in Racket 6.0. For earlier versions, need to `require` `racket/path`.

Commit 14003d0 has those changes and passes on Travis CI.

tfeb:
This smells like the right approach: I was going to do something which let you specify both where the root was and where the blog index was, with some defaulting, but I have forgotten how that code works and got as far as it puking with a contract violation in `serve/servlet`.

I do use `--serve` in fact, but only to make a new instance of the preview server without opening a new browser tab, not in production.

tfeb:
As of 14003d0afd20cf399895fef758c91d8c563cfcad this works really well for me: if I say `raco frog --root <my-root> -bp` the browser opens a tab on the root page for the blog and all the links work (so root is right).  So, for me, this is fine, and better than before!

greghendershott:
Thanks for letting me know. It's also working fine on my blog, which uses [a vanilla `.frogrc`](https://github.com/greghendershott/greghendershott.github.com/blob/master/.frogrc) with no special `prefix-uri` or `output-dir` -- so the changes didn't break that config.

So the interesting question is, should I merge this to master?

I'd talked about creating tests for permutations of configuration parameters. I haven't finished that, yet. In fact as I got into it, I started to feel that a truly effective test would require a "pile of files" for each configuration permutation -- not just a combo of values in `.frogrc`, but also things like `page-template.html` customized with certain links, and so on. At that point I felt discouraged, because if the tests require that much work -- and multiplied by all the permutations -- it's not realistic that people will keep the tests updated. Hopefully I'm just being dense and will be able to think of a more elegant way to go about it. But meanwhile, I don't have any spiffy new tests to run on these changes.

Also I'm going to be challenged for time for the next 4-5 days. I think I should probably hold off on doing the merge until after that, so that if anything does break, I'll be in a position to respond quickly.

tfeb:
Yes, I agree.  There is no urgency at all from my point of view: I would like not to have to maintain my own fork for ever if I can avoid it, but it makes no difference if it is merged now or in due course, so better to wait .

greghendershott:
I'll definitely merge to master in a matter of days, not weeks/months.

Meanwhile I pushed commit 0d3b9af adding to README and updating the `--root` usage text.

greghendershott:
Merged! Thanks again!!

tfeb:
Thanks!

PR #131, opened 2015-06-19
add www-offset for hosting in a subdirectory of the server

willghatch:
So I decided to switch from Jekyll to Frog for my (very small...) blog.  I host it on a virtual server that I rent, but I host other static stuff there too.  Therefore my blog is in a subdirectory (named "blog") so it can be updated and the directory can be blown away and replaced by the blog build process without affecting the other stuff there.  While switching to Frog I found that it didn't quite seem possible to set it up to do that, so I added that functionality.  The only downside I see is that it adds some /./ paths in some parts that are built with strings rather than path objects if you don't use an offset.  It's invisible to browsers and I personally don't mind it, but perhaps you will.  Let me know your thoughts.

(below is the commit message)

Basically this allows you to add extra path elements between the
output-dir/hostname and the actual blog contents.  It does not solve the
issue of copying over the static elements in the js, css, and img
directories, but they are hard coded in the templates anyway, so I
assume frog isn't really managing those anyway (and I simply added a
script to copy those over when needed in my setup process).

### DIFF
--- a/README.md
+++ b/README.md
@@ -696,6 +696,18 @@ The `--preview` flag should open on your blog's index page at
 `/blog/index.html`, automatically. (But there's also a `--root` flag
 in case you need to control it more specifically.)
 
+# MathJax
+
+To use MathJax:
+
+1. Add configuration to the `<head>` of your `page-template.html`. For
+   a standard MathJax configuration simply add `@math-jax[]` (call the
+   the `math-jax` function from `widgets.rkt`).
+
+2. In your markdown source files, use `\\( some math \\)` for inline
+   and `\\[ some math \\]` for display. (Note the _double_
+   backslashes, `\\`, because in markdown `\` already has a meaning.)
+
 # Bug reports? Feature requests?
 
 Please use [GitHub Issues][].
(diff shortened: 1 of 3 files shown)


willghatch:
Hey, thanks for adding the feature.  I appreciate it!

greghendershott:
Well, I want to thank you very much for submitting this pull request, which nudged me to address the original one from @tfeb. Again I'd like to apologize it took me so long to act.

willghatch:
Oh, no worries. I decided I wanted to switch to frog for various reasons, but that was the one blocking feature. So I just threw it together, and I'm happy to now know more about the internals of my blogging platform even if I'm not actually using the code I wrote anymore.

Also, if I had just looked at the existing pull requests before hacking away I would have noticed someone else already did it.

Thanks again.

On July 3, 2015 8:48:36 AM MDT, Greg Hendershott <notifications@github.com> wrote:
>Well, I want to thank you very much for submitting this pull request,
>which nudged me to address the original one from @tfeb. Again I'd like
>to apologize it took me so long to act.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/greghendershott/frog/pull/131#issuecomment-118366877

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

PR #137, opened 2015-07-11
Relocatable sites

khinsen:
This pull request combines two changes:
 - it makes `uri-prefix` available as a template variable
 - it inserts `@|uri-prefix|` at various places in `page-template.html`.

This makes Frog-generated sites easily relocatable. For a site hosted at `server/uri-prefix`, create a directory `uri-prefix`, move `css`, `img`, and `fonts` there, and in `.frogrc` set `uri-prefix` and `output-dir` to the same value. No need to change anything in `page-template.html`. Inside your content (Markdown/Scribble), use relative links.

### DIFF
--- a/example/_src/page-template.html
+++ b/example/_src/page-template.html
@@ -10,15 +10,15 @@
     <meta name="author"      content="@|author|">
     <meta name="keywords"    content="@|keywords|">
     <meta name="viewport"    content="width=device-width, initial-scale=1.0">
-    <link rel="icon"      href="/favicon.ico">
+    <link rel="icon"      href="@|uri-prefix|/favicon.ico">
     <link rel="canonical" href="@|full-uri|">
     @(when rel-next @list{<link rel="next" href="@|rel-next|">})
     @(when rel-prev @list{<link rel="prev" href="@|rel-prev|">})
     <!-- CSS -->
-    <link rel="stylesheet" type="text/css" href="/css/bootstrap.min.css">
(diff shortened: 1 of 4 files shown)


greghendershott:
That's a great idea! Merged. Thanks!

greghendershott:
p.s. I rebased to master before merging.

greghendershott:
p.p.s. I pushed a commit c29312f to simplify a few expressions. I really appreciate your time helping with this -- from the original issue, all the way through to this great PR. Which worked perfectly well, I just got nit-picky. :)  Thanks again!

tfeb:
@greghendershott / @khinsen:  thanks for putting this in, I was working my way around to suggesting something about half as good.

With this and the recent MathJax stuff, frog is really pretty cool I think: I spent way too long looking at things that either required me to run & maintain some vast blob of badly-written code in a language constructed from the bits of decayed food that accumulated in someone's keyboard, have my soul owned by Evil Megacorp, or both: frog just works, is implemented in a pleasant language, and also looks good.

khinsen:
@tfeb I came to Frog for much the same reasons, it strikes the right balance (for me at least) in terms of features and hackability.

@greghendershott Thanks for (1) Frog and (2) being so open to contributions!

greghendershott:
@tfeb Once upon a time I started to install something else and its requirements, felt it was too "heavy", and wrote Frog instead. On balance, definitely not a net time saving. :)  But glad I did.

@khinsen I really appreciate the contributions!

PR #139, opened 2015-07-16
Fix unicode titles in scribble pages

khinsen:
For reasons that I don't understand, Scribble parses strings with non-ASCII characters into lists of strings and isolated characters. This breaks `string-join` in `make-title` in `non-posts.rkt`. This may not be the best way to fix this, but at least it works.

### DIFF
--- a/frog/non-posts.rkt
+++ b/frog/non-posts.rkt
@@ -77,14 +77,19 @@
     (cons uri-path v)))
 
 (define (make-title xs path)
+  (define (x->m el)
+    (define m (xexpr->markdown el))
+    (if (char? m)
+        (string m)
+        m))
   (or (for/or ([x (in-list xs)])
         (match x
           ;; First h1 header, if any -- Scribble style with <a> anchor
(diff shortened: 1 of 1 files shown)


greghendershott:
Thanks for catching this, and providing a fix.

I can repro this if I put a `λ` in the `@title{}` of the example project's `_src/Example-Non-Post-Scribble-Page.scrbl`,

I think the real problem is actually down in `xexpr->markdown`. That should always return a `string?`. In fact I should probably add a contract to make that explicit:

```
(define/contract (xexpr->markdown x [block-suffix ""])
  (->* (xexpr/c) (string?) string?)
```

And to make it honor its own contract, this clause:

```
    [(? valid-char? i) (integer->char i)]
```

should actually be:

```
    [(? valid-char? i) (string (integer->char i))]
```

Since I've already worked this out in the code, I think I'll just push that as a commit myself to close the issue. But thanks again for providing a reasonable pragmatic fix!

PR #143, opened 2015-08-18
Allow --edit after a -N call

ianj:
I found that often I create a new post but forget to supply --edit. This allows one to reuse the -N command with an --edit flag to not rewrite the file, but simply open the difficult to remember/stubborn-to-tab-complete file name in the editor.

### DIFF
--- a/frog/new-post.rkt
+++ b/frog/new-post.rkt
@@ -67,13 +67,16 @@ EOF
                                 our-encode)
                             extension))
       (define pathname (build-path (src/posts-path) filename))
-      (when (file-exists? pathname)
-        (raise-user-error 'new-post "~a already exists." pathname))
-      (display-to-file* (format template
-                                title
-                                (date->string d #t)) ;do include time
+      (cond
+        [(file-exists? pathname)
+         (unless (enable-editor?)
+           (raise-user-error 'new-post "~a already exists." pathname))]
+        [else
+         (display-to-file* (format template
+                                   title
+                                   (date->string d #t)) ;do include time
                         pathname
-                        #:exists 'error)
+                        #:exists 'error)])
       (displayln pathname)
       (when (enable-editor?)
         (system (editor-command-string 


greghendershott:
Good idea -- thanks!

greghendershott:
p.s. A small request for future commits, if any: I prefer commit messages line wrapped in the customary way. Anyway thanks again for submitting this!

ianj:
Understood. First time using github's pencil-in feature. I realize that this feature is limited in that it requires the --edit -N <title> call to be on the same date as the original -N call. I'll have to extend that to

1. open the latest post of that title if there are no other posts with that title
2. if there are other posts, find all files that have been generated on whatever day from that same title text, then list them, numbered, reverse chronologically. If given an extra argument, say --select N, open the Nth latest post with that title.

greghendershott:
By the way, I don't use the `--edit` feature myself; it orignated from a pull request.

I run Frog in an Emacs `*shell*` buffer. The `-n` and `-N` commands output the full pathname of the file just created. So I just copy that text, choose `find-file`, and paste that into the minibuffer; done. Probably there's a moral equivalent for most other combos of terminals and editors?

Likewise to edit among N post sources, again in Emacs I just open a directory buffer on `_src/posts`, incremental search, and voila. Or, since I use `ido`, I can do an incremental/fuzzy search directly from the `find-file` prompt.

I liked your original PR because I understood it to be a simple, harmless way to help a user with a mistake. What you're describing now seems less simple, and more like it could (should?) be done external to Frog. But that's just my first impression, so I don't mean to discourage you if you think it would really add value.

ianj:
I'm treating frog like more of a database than it is. I spent all my writing time tonight into this over-engineering. I feel like frog should give more info about what its posts are and what its posts aren't, so an actual database might be prudent rather than yet another feature-starved flat file "database" that I was writing.

greghendershott:
Just my perspective: Frog is file oriented. There are tools to work with files in database-ish ways, including as if they were rows in a database.

If grep-ing the {.md,.scrbl} files for the metadata (title, date, tags) feels too gross, perhaps we could add a flag to Frog to print the info to stdout in some format (sexprs, CSV, JSON) that would be handy for other tools?

Oh wait. There's already a `.frog/build` cache file. Would that help you?

greghendershott:
To be clear, this was created in response to @samth urging me to make Frog builds even faster. Currently it's intended to be "private", no guarantees.

greghendershott:
It's just a `racket/serialize` dump of the `listof` `post` `struct`s.

PR #144, opened 2015-08-19
Add a parameter to disable showing parent tweets when embedding tweets

bmastenbrook:
["No great deed was committed here"](https://brian.mastenbrook.net/blog/2015/we-considered-ourselves-to-be-a-humorous-people.html)

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,6 +261,10 @@ feed-image-bugs? = true
 #
 auto-embed-tweets? = true
 
+# When embedding tweets that are replies, show the parent tweet along
+# with the reply?
+embed-tweet-parents? = true
+
 # Try to automatically link symbols in Markdown ```racket fenced code
 # blocks, to Racket documentation?
 racket-doc-link-code? = true
--- a/example/.frogrc
+++ b/example/.frogrc
@@ -67,6 +67,10 @@ feed-image-bugs? = true
 #
 auto-embed-tweets? = true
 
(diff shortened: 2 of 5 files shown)


bmastenbrook:
By the way, that's not a comment on frog, just a link to the example showing why I needed this :-)

greghendershott:
Thanks! Could you please also push a commit updating both:

- [example .frogrc](https://github.com/greghendershott/frog/blob/master/example/.frogrc) (this is also used by `raco frog --init` when people create a new blog project)

- [.frogrc section of README.md](https://github.com/greghendershott/frog#per-project-configuration-frogrc) (because documentation)

greghendershott:
Also, I notice the Twitter query parameter is named `hide_thread`. Thread not parent. Does it merely hide the parent (previous) tweets, or the entire thread including child (following) tweets?

If the latter, do you think maybe the Frog parameter should be named more like the Twitter query param, talking about "threads" not "parents"?  Such as `current-embed-tweet-threads`?

(Maybe I'm being nit-picky but I wanted to ask.)

bmastenbrook:
The Twitter parameter is named "hide_thread", but the embedding only picks up the parent tweet as far as I can tell. I can change the name if you'd like, but the name for the parameter in Twitter's API didn't really make a whole lot of sense as it was to me.

greghendershott:
In that case the name sounds perfect.

Thanks also for commit with the docs.

Issue #162, opened 2016-04-30
Multiple authors?


bennn:
Can frog currently:
- list the author of a blog post on the post
- allow multiple authors (without changing the `.frogrc`)?

If not, I'm thinking I'll use tags like "by John Doe" as a hack. But it would be nice to start posts with:

```
  Title: ...
  Author: ...
  Date: ...
  Tags: ...
```

(My use case: I'd like to use frog for an "organizational" blog. Most posts will have different authors and I'd like to give each author credit.)

gasche:
Indeed, I was just looking for the same feature this morning. It would be really nice to have.

gasche:
So I started sending preliminary commits going in the direction of Author handling (but it's still work-in-progress) at #163, any feedback is welcome.

gasche:
Thanks a lot!

PR #163, opened 2016-05-04
WIP towards authors handling

gasche:
I'm trying to add multi-author handling to Frog ( feature request #162 ) , but this is my first Racket hacking experience so I'm going rather slow at it. I would welcome feedback along the way: I'll push my patches when I feel they are reasonable, and rebase the PR when I have more to send -- or to make changes according to comments.

The very first patch in this PR (and for now the only one) is just changing the internal implementation of metadata parsing, with no observable change of behavior (I hope). The metadata header is parsed into an associative list.

I then tried to move the handling of mandatory fields from `meta-data` to `read-post` (my idea was to have `meta-data` be agnostic over the specific values of the metadata header, and move the mandatory/optional logic to `read-post`), but the current tests (which were very useful to me, thanks) expect `meta-data` to do the mandatory-fields checking. I'm interested in opinions on what is best: I could remove the tests and defer the mandatory-fields checking to `read-post`, or I could go the less invasive route of just adding `header` as an extra value returned by `meta-data`, so that mandatory fields are still handled there but `read-post` can extra optional fields (such as `Author`).

### DIFF
--- a/frog/posts.rkt
+++ b/frog/posts.rkt
@@ -106,6 +106,26 @@
           more?
           (~> body enhance-body xexprs->string))))
 
+(define (read-meta-data-line input)
+  (match input
+         [(pregexp
+           (string-join (list
+                         "^\\s*([^:]+)\\s*"; key (space-trimmed)
+                         ":\\s*([^\n]+)\\s*"; value (space-rimmed)
+                         "(?:\n(.*))?$"; rest (optional)
+                         ) "")
+           (list _ key val rest))
+          (cons (cons key val) rest)]
+         [_ #f]))
+
+(define (read-meta-data-acc acc input)
+  (match (read-meta-data-line input)
+    [(cons binding rest)
+     (read-meta-data-acc (cons binding acc) rest)]
+    [#f (values (reverse acc) input)]))
+
+(define (read-meta-data input) (read-meta-data-acc '() input))
+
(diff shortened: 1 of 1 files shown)


gasche:
Ping? I haven't worked at it since this first PR, and I must say that some sort of feedback would help going forward.

greghendershott:
@gasche Thank you very much for submitting this pull request. You caught me during a period where I've been flat-out at $NEW_DAY_JOB. Furthermore, I haven't been working on Frog for some months, so, I kept failing to find time to reload my brain enough to evaluate this and give any useful feedback.

It looks like @LeifAndersen has helped and made a new PR #164 building on your commit. If I understand correctly, I should close this PR and focus on that one @gasche ?

LeifAndersen:
Ya, you should. The two of us work in the same lab and he has told me multiple times (in person) that I should finish this for him because he wouldn't have time for it for a bit.

greghendershott:
@LeifAndersen OK, will close this one.

@gasche Thank you again for submitting this. I apologize for being incommunicado so long.

gasche:
Yep, that's fine. Thanks to you and @LeifAndersen for working on this!

For the record, the reason why we need multiple-authors handling is that we're setting up a [research lab blog](http://prl.ccs.neu.edu/blog/) using Frog.

greghendershott:
As discussed on #164 addressed this in commit 1d99938.

Thank you @gasche and @LeifAndersen !

PR #164, opened 2016-06-24
Support blogs that wish to have multiple authors

LeifAndersen:
Finishing up the work in [this pull request]{https://github.com/greghendershott/frog/pull/163}, I have added the ability for Frog blogs to support multiple authors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -351,7 +351,10 @@ feeds. A "Continue reading..." link is provided instead.
 
 `Date` must be an ISO-8601 datetime string: `yyyy-mm-ddThr:mn:sc`.
 
-`Tags` are optional (although you have to include the `Tags:` part).
+`Tags` are optional.
+
+`Authors` are also optional for multi-author blogs, and work similarly
+to tags.
 
 > The tag `DRAFT` (all uppercase) causes the post `.html` file _not_
 > to be generated.
--- a/example/_src/index-template.html
+++ b/example/_src/index-template.html
@@ -1,7 +1,10 @@
 <article>
   <header>
     <h2><a href='@|uri-path|'>@|title|</a></h2>
-    <p class='date-and-tags'>@|date| :: @|tags|</p>
+    <p class='date-and-tags'>@|date|
+      @(if (equal? tags "") "" (format " :: ~a" authors))
+      @(if (equal? tags "") "" (format " :: ~a" tags))
+    </p>
   </header>
   @|content-only|
   @(when more?
--- a/example/_src/post-template.html
+++ b/example/_src/post-template.html
@@ -1,7 +1,10 @@
 <article>
   <header>
     <h1>@|title|</h1>
-    <p class='date-and-tags'>@|date| :: @|tags|</p>
+    <p class='date-and-tags'>@|date|
+      @(if (equal? tags "") "" (format " :: ~a" authors))
+      @(if (equal? tags "") "" (format " :: ~a" tags))
+    </p>
   </header>
   @|content|
   <footer>
--- a/frog/bodies-page.rkt
+++ b/frog/bodies-page.rkt
@@ -81,6 +81,7 @@
        (a ([href ,(atom-feed-uri "all")])
           (img ([src ,(canonicalize-uri "/img/feed.png")])))))))
 
+
 (define (tags-list-items)
   (for/list ([(k v) (in-dict (tags-alist))])
                `(li ,(tag->xexpr k))))
--- a/frog/feeds.rkt
+++ b/frog/feeds.rkt
@@ -59,7 +59,8 @@
    (display-to-file* file #:exists 'replace)))
 
 (define (post->atom-feed-entry-xexpr tag x)
-  (match-define (post title src-path modtime dest-path uri-path date older newer tags blurb more? body) x)
+  (match-define
+    (post title src-path modtime dest-path uri-path authors date older newer tags blurb more? body) x)
   (define item-uri (full-uri/decorated uri-path #:source tag #:medium "Atom"))
   `(entry
     ()
@@ -114,7 +115,8 @@
    (display-to-file* file #:exists 'replace)))
 
 (define (post->rss-feed-entry-xexpr tag x)
-  (match-define (post title src-path modtime dest-path uri-path date older newer tags blurb more? body) x)
+  (match-define
+    (post title src-path modtime dest-path uri-path authors date older newer tags blurb more? body) x)
   (define item-uri (full-uri/decorated uri-path
                                        #:source tag #:medium "RSS"))
   `(item
--- a/frog/frog.rkt
+++ b/frog/frog.rkt
@@ -8,7 +8,6 @@
          racket/path ;moved to racket/base only as of Racket 6
          racket/runtime-path
          racket/set
-         racket/hash
          rackjure/threading
          web-server/dispatchers/dispatch
          web-server/servlet-env
@@ -207,14 +206,23 @@
   (define new-posts
     (fold-files* on-file (make-hash) (src/posts-path) #f))
   ;; Make hashes of tags -> posts, old and new.
-  (define old-tags (let ([ht (posts->tags/authors old-posts)])
-                     (hash-union! ht (posts->tags/authors old-posts #:type 'authors)
-                                  #:combine (λ (a b) a))
-                     ht))
-  (define new-tags (let ([ht (posts->tags/authors new-posts)])
-                     (hash-union! ht (posts->tags/authors new-posts #:type 'authors)
-                                  #:combine (λ (a b) a))
-                     ht))
+  (define (merge-tags a b)
+    (define ret (make-hash))
+    (for* ([(k v) (in-hash a)]
+           [(k* v*) (in-hash b)]
+           #:when (equal? k k*))
+      (hash-set! ret k v))
+    (for ([(k v) (in-hash a)]
+          #:unless (hash-has-key? b k))
+      (hash-set! ret k v))
+    (for ([(k v) (in-hash b)]
+          #:unless (hash-has-key? a k))
+      (hash-set! ret k v))
+    ret)
+  (define old-tags (merge-tags (posts->tags/authors old-posts)
+                               (posts->tags/authors old-posts #:type 'authors)))
+  (define new-tags (merge-tags (posts->tags/authors new-posts)
+                               (posts->tags/authors new-posts #:type 'authors)))
   (all-tags new-tags)
   ;; Make a list of the post source paths, sorted by post date.
   ;; (This is needed by index pages, which list posts ordered by date.)
--- a/frog/frog.rkt
+++ b/frog/frog.rkt
@@ -8,6 +8,7 @@
          racket/path ;moved to racket/base only as of Racket 6
          racket/runtime-path
          racket/set
+         racket/hash
          rackjure/threading
          web-server/dispatchers/dispatch
          web-server/servlet-env
@@ -206,8 +207,14 @@
   (define new-posts
     (fold-files* on-file (make-hash) (src/posts-path) #f))
   ;; Make hashes of tags -> posts, old and new.
-  (define old-tags (posts->tags old-posts))
-  (define new-tags (posts->tags new-posts))
+  (define old-tags (let ([ht (posts->tags/authors old-posts)])
+                     (hash-union! ht (posts->tags/authors old-posts #:type 'authors)
+                                  #:combine (λ (a b) a))
+                     ht))
+  (define new-tags (let ([ht (posts->tags/authors new-posts)])
+                     (hash-union! ht (posts->tags/authors new-posts #:type 'authors)
+                                  #:combine (λ (a b) a))
+                     ht))
   (all-tags new-tags)
   ;; Make a list of the post source paths, sorted by post date.
   ;; (This is needed by index pages, which list posts ordered by date.)
@@ -284,7 +291,8 @@
       (not (equal? post (hash-ref old-posts (post-src-path post) #f)))))
   (define (post-has-tag? tag post)
     (or (equal? tag "all")
-        (member tag (post-tags post))))
+        (member tag (post-tags post))
+        (member tag (post-authors post))))
   (for ([(tag paths) (in-hash new-tags)])
     (define posts-this-tag
       (filter values
@@ -295,7 +303,7 @@
               (stale/posts? posts-this-tag)
               (stale/templates? tag))
       (write-stuff-for-tag tag posts-this-tag)))
-
+  
   ;; [3] Save `new-posts` (to be used as the `old-posts` for our next
   ;;     build).
   (serialize-posts new-posts)
@@ -333,17 +341,19 @@
   (not (member "UNLINKED" (post-tags p))))
 
 ;; Given a (hash/c path? post?) return a (hash/c string? (set/c
-;; path?)), where the string is a tag. In other words this returns a
-;; hash that maps a tag to all the posts that are marked with the tag.
-(define/contract (posts->tags posts)
-  ((hash/c path? post?) . ->  . (hash/c string? (set/c path?)))
+;; path?)), where the string is a tag or author. In other words this returns a
+;; hash that maps a tag or author to all the posts that are marked with the tag or author.
+(define/contract (posts->tags/authors posts #:type [type 'tags])
+  (((hash/c path? post?)) (#:type (or/c 'tags 'authors)) . ->*  . (hash/c string? (set/c path?)))
   (define h (make-hash))
   (for ([(path post) (in-hash posts)])
     (when (linked-post? post)
-      (for ([tag (in-list (cons "all" (post-tags post)))])
+      (for ([tag (in-list (cons "all" ((if (equal? type 'tags) post-tags post-authors)
+                                       post)))])
         (unless (equal? tag "")
           (hash-update! h tag (lambda (v) (set-add v path)) (set))))))
   h)
+
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
--- a/frog/params.rkt
+++ b/frog/params.rkt
@@ -25,6 +25,7 @@
 (define current-index-full? (make-parameter #f)) ;index pages: full posts?
 (define current-feed-full? (make-parameter #f))  ;feeds: full posts?
 (define current-show-tag-counts? (make-parameter #t))
+(define current-show-author-counts? (make-parameter #t))
 (define current-max-feed-items (make-parameter 999))
 (define current-decorate-feed-uris? (make-parameter #t))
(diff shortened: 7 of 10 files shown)


greghendershott:
Thank you for helping with this.

Do you understand why the tests are failing?

LeifAndersen:
Oops, odd, they were working on my machine, checking it out now.

LeifAndersen:
Okay, I found the reason why they are crashing. I use the `racket/hash` library (for `hash-union!`), which wasn't added until 6.3.

Before that, `hash-union!` was located in the unstable lib.

So it looks like we have 3 options:

1. Abandon the use of `hash-union!` and implement it ourselves.

2. Use `hash-union!` and have two versions of frog, one for older versions of racket and one for newer ones.

2a. This can be done statically using the package server's version exceptions

2b. Or dynamically using dynamic require to require the correct hash-union! function, based on the current Racket version.

3. Drop support for versions of Racket before 6.3. (Yuck)

I hate option 3. I absolutely would like frog to keep supporting versions of Racket going back to 6.0.  Option 1 works, and wouldn't be too much work to implement, same with option 2b. 2a could work (it'd be up to you to add the separate version exceptions), but this likely means that the version of frog for older versions of Racket are much less likely to get updates.

Which way would you prefer?

greghendershott:
Thanks for looking into it. I want to support older Rackets, and I'd rather not go down the road of version exceptions.  So I prefer the first option: Pretend Racket 6.3 doesn't exist yet, and write some equivalent of `hash-union!`. That's what I've tended to do in similar cases. If that's not too difficult in this case?

LeifAndersen:
Yup, I agree. (For the record, option 2b would also work here, because `hash-union!` 'did' exist, just in the unstable lib), but ya, in this case option 1 works just fine.

LeifAndersen:
Okay, I have removed the use of `racket/hasah`, hopefully it builds now...

LeifAndersen:
@greghendershott It looks like all of the tests pass now.

greghendershott:
Thanks. I'm reviewing the code more carefully now because I'm a little confused about the mixing of tags and authors. I think you're doing this to avoid duplicating code/logic in `build`, which is understandable.

But I'm wondering if it's OK for the `all-tags` parameter to get values for authors as well as tags. I'm not sure yet. (It would help if I'd worked on my code more recently.) Had you considered this?

LeifAndersen:
Eep, no, I didn't realize that existed. Ya, I'm not sure about that.

And yes, I did that to prevent having to duplicate the code for tags and authors.

I did consider having separate code for authors, but realized that doesn't work when you get to the feeds area, because that doesn't have `tags` in the URL.

greghendershott:
Well also I think there's a good intuition here: Tags and authors are very similar. You could even say that authors are a special case of tags, but distinguished in posts and in feeds. (Or that both are a special case of some other thing.)

Each tag gets its own index page and feed. Should authors, too?  Curious if you think the answer should be "Yes!" / "No!" / "don't really care".

greghendershott:
To put this another way:

Today people can use tags like "by John Doe" as a work-around (as mentioned in #162).

Imagine a lazy, cheap hack where tags starting with `"by "` get additional special treatment:

-  being put in an `authors` template variable (as you did) instead of `tags` template variable
-  being used as the `author` element in an ATOM feed
- one or two other things?

Otherwise, authors behave exactly like tags (e.g. they get their own index pages and feeds), and let's say this is desirable.

On the one hand, clearly it's a cheap hack to use the English text `"by "` as a magic prefix. :)

On the other hand, maybe in spirit this is close to the best way to do it?

LeifAndersen:
So, I absolutely agree that authors are a special type of tag. Or rather, I think authors and tags are part of something more general. I, however, don't think that doing `by Foo Bar` is a good convention and would rather have an `Authors` field. But I wouldn't complain too much if that's what you ended up wanting to do.

greghendershott:
I definitely want the front-end source to add a distinct Authors meta-data
field.

I was thinking about the implementation. I started to sketch out a
variation on your PR (taking it a step forward and to the side, so to
speak). But unfortunately I'm traveling for work this week and will need to
punt until at least Friday.

LeifAndersen:
Ah, okay. That's fair.

Ya, I kind of think the implementation should put them together, but I'm not exactly sure of the best way to do it, and so that way seemed like the least terrible kludge in terms of being easy to write and gets authors.

However I completely agree that that implementation is a total smell, and could be implemented much better.

eestrada:
This is a related comment, albeit a bit of a direction change. With the metadata in the markdown, I started looking at the Python MarkDown package and metadata parsing is actually [relegated there as an extension](http://pythonhosted.org/Markdown/extensions/meta_data.html), so that metadata can be any arbitrary key/value pair and be used with any code that depends on MarkDown in general.

I think that may be a good separation of concerns here as well. Since @greghendershott heads up both [frog](https://github.com/greghendershott/frog) and [MarkDown](https://github.com/greghendershott/markdown), it may be good to move general metadata parsing over to MarkDown while work is being done on the "Authors" metdata already. Or maybe the can is kicked down the road on that as a plan for a future addition to MarkDown. Anyway, it seems like metadata parsing is useful for MarkDown outside of frog alone and makes more sense living directly there in my eyes.

gasche:
Ping? Having support for multiple authors on a single blog is something that we really need.

gasche:
Ping ? It would really be nice to have more freedom on the authors of a blog post.

I looked at the implementation in this PR and it seems that the `authors` metadata does not in fact flow to the `author` feed in the generated syndication feed. If I'm not mistaken, then would it maybe be possible to fix this? Having authors marked in the syndication feed is a feature missing from tag-style emulation.

gasche:
I am sorry to be a bit heavy-handed on this issue, but for multi-author blog post it would be *really* important to have author names actually show up properly in the syndication feeds. (I'm not sure this is actually part of this PR, but it seems fairly simple to build on top.) If we cannot get this issue to move forward, I will try to convince my colleagues to use a fork of frog; if it does not work, I will switch our blog to another blog generator.

I understand the agony over the design choices -- I'm the same with other software. But having a maybe-not-optimal design today does not restrict your ability to use a better design later (in fact, it may make it easier to get experience with the current proposal and find the better design from that). We spent 9 months without the feature by now, and this hasn't made the design of Frog any better -- at a net loss to our blog.

eestrada:
@gasche I started playing with this last night (since I would like to see it available as well), and I think I am close to figuring it out. The work is currently all local to my machine.

I have been installing frog from @LeifAndersen's fork to get Author support, and it is fully functional as is. Once I get multiple author support into the feeds, I will create my own Github fork off of @LeifAndersen's fork and possibly open a new pull request here.

After searching the interwebz about the RSS spec it doesn't seem like there is author support of any sort, multiple or otherwise. So multiple author support will probably only be available in Atom feeds. What I have already works, but it is strangely doubling the authors, so I need to look at that and figure out what I broke. Hopefully it won't take me more than a day or two. Then you will at least have a fork of frog with full support for multiple authors until @greghendershott has more time to dedicate to reviewing pull requests here on the frog repo.

gasche:
Thanks! RSS's `<item>` elements have an optional `<author>` subelement. The specification asks for an email address, but gives examples of the form `email@example.com (Firstname Lastname)`, so it is actually more free-form; I would recommend simply reusing the content of the `Author` metadata field, delegating the email-or-not choices to the blog articles.

To be clear, I think this is not really a case of @greghendershott not having any time for Frog -- it is not unmaintained software. I have reported a couple of issues / submitted a couple simple PRs and they were adressed/merged quickly. I don't know Greg, but from my own experience as a (not generally very good) maintainer, I would rather see a case of perfectionism at play: I'd look at some PRs, think about the *nice* way to do it, but it's more work and I don't have the time and end up not doing anything on this maintenance cycle -- rinse and repeat. I appreciate the work that was and still is put into Frog, it's just that multi-author support is a must-have feature for a multi-author blog.

eestrada:
@gasche Can you post a link to the RSS specification? Finding the Atom spec was pretty easy for me. Less so with the RSS 2.0 one.

gasche:
Sure, the reference I use is [this one](https://cyber.harvard.edu/rss/rss.html), in particular the [Elements of `<item>`](https://cyber.harvard.edu/rss/rss.html#hrelementsOfLtitemgt) section.

greghendershott:
Hi. For the past year+ I've been working full-time outside of Racket (sadly) -- with not much leftover time and energy. I've tried to keep up with bugs. But I've been scared to make changes that could create problems that I don't have time to fix. i.e. "Do no harm" mode.

Plus then more recently I learned that http://blog.racket-lang.org/ had started to use Frog -- making me even more conservative and concerned about breakage. (Incidentally, the Racket blog has multiple authors, and they didn't ask for this, so that weighed into the mix, too.)

@gasche 
> I'm interested in eventually having "the best way to do it" in Frog, but I'm also interested in having any acceptable way to do it (as in: the author name shows up in the expected place in the syndication feed) 

In that case hasn't there been available all along a self-help hack like:
1. Use `"by Some Name"` plain tags for authors
2. Write your own utility to alter the RSS and/or Atom feed files
3. Run your utility after `raco frog -b`
?

greghendershott:
So, I just spent some time reviewing the original issue #162, PR #163, and the comment thread here on PR #164.

It looks like the actual requirements grew -- or at least got more specific -- over time.

1. Originally this feature was about letting people (a) view posts by-author, and (b) having a feed for each author containing their posts.  All of which _can_ be handled already by using `"by Author Name"` plain old tags.

2. The PR as it stands now only does the above -- only does what you could already do with plain tags. If I'd clicked the `Merge` button, that's all we'd have.

3. Not until [Sep 2016](https://github.com/greghendershott/frog/pull/164#issuecomment-250771487) is there mention of writing things _inside_ Atom and RSS feeds.

4. And only in the last 24 hours does it get specific what that would mean in the Atom and RSS formats.

So on the one hand I still feel bad I haven't had more time -- maybe I could have teased this out or realized it on my own, much sooner.

On the other hand I feel fine that I trusted my gut and didn't just click `Merge`.

---

What next? Here's a proposal in the spirit of "what's the simplest thing that could possibly work":

- If a plain old tag is of the form `by <any text>`, we write special author info in the Atom and XML feed files.

- That's it!

(Yes this means a tag like `"by the water"` would result in `"the water"` being treated as an author. I'm going to shrug and say OK, that could happen; too bad.)

Hopefully @eestrada is willing to help pin down the exact Atom and RSS changes required. Either he or I could code this in a new PR.

What do you think?

gasche:
I think that's conceptually uglier than the current proposal, but I am also totally fine with it. It would be nice if multiple-author posts were supported, so the author info should collect all `by .*` tags and concatenate their payload with commas as separator.

P.S.: I don't particularly care for the ability to search for all posts by someone personally, although it is a nice-to-have.

greghendershott:
Speaking of multi-author posts, do you know the correct thing to do in both Atom and RSS feeds?

May we simply include multiple `<author>` elements?

e.g. RSS
```xml
<item>
 <author>"Alice Doe"</author>
 <author>"Bob Doe"</author>
 ....
</item>
```

e.g. Atom
```xml
<item>
 <author><name>"Alice Doe"</name></author>
 <author><name>"Bob Doe"</name></author>
 ....
</item>
```

greghendershott:
Oh, I think I read your previous comment too quickly. I think you're suggesting that it must be one `<author>` element, and we must concat with commas. Correct?

gasche:
I have used comma-separated author names in a single `<author>` tag for another blog in the past and it worked well. I think having multiple items may work less well (not be supported by some readers, etc.), but I have not tested.

greghendershott:
[This](http://stackoverflow.com/questions/5855993/multiple-authors-in-rss-or-atoms) implies:

- RSS should be single `<author>` element.
- Atom may be multiple `<author>` elements, however some consumers will ignore all but one.

This is starting to job my memory, with feeds, how big a gap there is between (partial) specifications and real-world practice.

Anyway @gasche you seem to have a specific use case in mind, so we may as well start by making it work for your subset of the real-world. :smile:

greghendershott:
OK I'll stick with single element, comma-separated, for both RSS and Atom.

gasche:
The use-case I have in mind is that the readers of the blog should be able to know, as they read a post, who the author is. It is specific, but also rather canonical within the possible meanings of "multi-author support".

greghendershott:
I mean your use-case for _feeds_. If you have some specific feed-reader in mind that will expect multiple authors in a format, and I don't do it that way, why am I even.

greghendershott:
@gasche @eestrada @LeifAndersen When you have time, could you please take a look at commit 76128b0 on the `authors` branch and let me know if that will work for you?

eestrada:
Using a custom prefix works for me if that is the fastest path to meeting @gasche's needs right now. I think down the line it would help for moving MarkDown files between blogs to have `Authors` in the frontmatter as a separate piece of metadata. I know Pelican lists multiple Authors this way. I am guessing other static site generators probably do too.

In case anyone is curious, I forked the repo from @LeifAndersen and pushed my changes for the sake of reference. The change is trivially small. But I hope to contribute more in the future when I get the chance, so having a github fork makes sense anyway. [Here is the branch](https://github.com/eestrada/frog/tree/authors) and [here is the specific commit that adds support for Atom](https://github.com/eestrada/frog/commit/1743fde73f768bc8debcdfbd35698a913652cdf0).

eestrada:
@greghendershott @gasche I just had an idea that could meet the needs of both compliant and non-compliant feed readers:

* List Authors as separate entities in the feeds
* List the last Author as a mashup of all of them

Like so:

```xml
<item>
 <author>"Alice Doe"</author>
 <author>"Bob Doe"</author>
 <author>"Alice Doe, Bob Doe"</author>
 ....
</item>
e.g. Atom

<item>
 <author><name>"Alice Doe"</name></author>
 <author><name>"Bob Doe"</name></author>
 <author><name>"Alice Doe, Bob Doe"</name></author>
 ....
</item>
```

That way non-compliant readers (which to my understanding tend to only pick up the last author entity) will work, and compliant readers will pick up multiple as separate entities. There is some redundancy, obviously, but I think it could be the best compromise of both. In previous comments both @greghendershott and @LeifAndersen weren't very eager to merge this PR because of some implementation details that they didn't like, so maybe it is just better to use @greghendershott's "by " prefix tag parsing commit for the time being. I would be sad to see the `Author` metadata go though.

eestrada:
Just realized why I was getting duplicate authors: I hadn't deleted the old code, just added new; I was very tired the other night. Now nemoved the old authors line, rebased and force pushed to my fork. [Here is the updated commit](https://github.com/eestrada/frog/commit/26149eb7f3ee8b1f657e820495e319be0d18cc77).

gasche:
I realize I may have been unclear in the past, sorry: I have no evidence of feed readers barfing multiple `<author>` items if this is what the specification recommends, I am just saying that I've used a single comma-separated field in the past and that it worked. It would be totally reasonable to comply with the specification, and wait until someone using the feature has a multi-author post to maybe notice a problem and change it then.

greghendershott:
@eestrada 

> I just had an idea that could meet the needs of both compliant and non-compliant feed readers:
>
> - List Authors as separate entities in the feeds
> - List the last Author as a mashup of all of them

Good idea.

> In previous comments both @greghendershott and @LeifAndersen weren't very eager to merge this PR because of some implementation details that they didn't like, so maybe it is just better to use @greghendershott's "by " prefix tag parsing commit for the time being. I would be sad to see the Author metadata go though.

Having slept on this I'm a little sad about it, too.

Internally I prefer these author annotations to be _implemented_ as a (tiny) specialization of tags, as opposed to some whole other thing.

Externally I do like the idea of explicit `Author:` metadata a.k.a. "front matter" as with other blog tools.

Tell you what, I'd like to take another pass at this, to do both. Best case it turns out well and we can merge that. Worst case we can merge the `authors` branch I pushed yesterday. To be clear I'm taking in terms of taking another day or two, not weeks or months!

greghendershott:
OK I think I'll be ready to push another branch tomorrow. This uses `Author:` metadata. However internally these are handled as qualified tags, providing the mostly common path I really wanted.

greghendershott:
I went ahead and pushed this as commit e8e1c2e on a new `authors-metadata` branch. This was branched off `authors`, so the commit reflects the delta from that. Obviously I'll squash this before merging. In fact if it's too confusing to review as-is, I could go ahead and squash and create a new PR; let me know if that would help.

greghendershott:
Pushed another commit to make this work on older Rackets.

Commits aren't yet squashed, but here is the [net diff from `master`](https://github.com/greghendershott/frog/compare/e1c0c48...819f4c6).

eestrada:
There is a lot I like about the new `authors-metadata` branch. I like that authors are just a special case of tags. I like that the last author item in a feed is all the authors together (yay).

Here is the one thing I dislike: that authors are marked as different by pre-pending a value to their actual string data. It smells bad to me for some reason.

It would take more refactoring, but this is how I would probably do it: instead of `tags` being a plain `list`, make it an `alist` instead (or a `hash`, it doesn't really matter). The tags themselves would be the keys, and their type as the value (probably a symbol). It would be like light metadata. Plain tags could have a type of `#f`. So it could look something like this:

```racket
(list ("foo" . #f) ("bar" . #f) ("Alice Baker" . 'author) ("Charlie Dean" . 'author))
```

That way, the tag data itself moves through the pipeline untouched. This would also open the door to other specialty tags types in the future if needed. @greghendershott, tell me what you think. Good idea? Bad idea? Tinfoil hat idea (i.e. I am being paranoid over something that isn't really an issue)?

I'd be happy to take a stab at implementing it if you deem it a good idea.

greghendershott:
Clean sheet? I would probably do that -- or another equivalent of a `union` or ADT in Racket, like:

```
(struct tag (text)) 
(struct plain-tag tag ())
(struct author-tag tag ())
```

And use `tag-text` generally, and check `plain-tag?`/`author-tag?` or `match` on `plain-tag`/`author-tag` when that mattered.

In fact those were the first lines of code I started to write on Friday. But I backed off because I realized:

- I haven't looked at Frog's code in quite a few months; not sure how many things this would break.
- @gasche is understandably eager to have something good-enough, soon, and so am I.

Someday Frog should probably be deeply overhauled in a number ways, but meanwhile.

eestrada:
@greghendershott I agree. Like I said, it would take more refactoring. What I meant (and should have said) was, "It will take more refactoring than there is probably time for right now." I agree that pre-pending "Author: " to the tag string is probably the fastest way to getting this working right now. As I also said, I like everything else about the commit.

The :shipit: squirrel says "Ship it!" and so do I. I will separately take a stab at trying tags as classes, or attaching separate metadata to the tag metadata (recursive metadata!) and see if I can keep it general. If I find something nice and general, I will open up a merge request. Otherwise, it will just be an exercise in design and Racket hacking.

greghendershott:
Good! I'll aim to squash and merge tomorrow (when I have a fresher brain to take one last look).

greghendershott:
As discussed, closing in lieu of commit 1d99938 just merged to `master`.

Thank you @LeifAndersen@eestrada and @gasche!

PR #176, opened 2016-12-20
Fix bad syntax highlighting

### LINKED ISSUE #165: No syntax highlighting for scribble posts.
There is no syntax highlighting for posts written in scribble.

In this case we don't even really need pygments to do the highlighting, as scribble's rendering engine can already do syntax highlighting. (Although I don't know how much of that you use and how much of it you re-built.)

But it would be nice to have syntax highlighting in my code when written in scribble.

sorawee:
Suppose the text passed to `pygment-code` is following:

```
-  RPROMPT='${editor_info[overwrite]}%(?:: %F{1}⏎%f)${VIM:+" %B%F{6}V%f%b"}'
+  RPROMPT='[%D{%L:%M:%S %p}]'
+  #RPROMPT='${editor_info[overwrite]}%(?:: %F{1}⏎%f)${VIM:+" %B%F{6}V%f%b"}'
```

Obviously, we should be able to highlight it. However, `texts` will not be all string. For example, `⏎` will result in a code (number) 9166. `'` will result in `'rsquo`. Therefore, the `match` fails, and it falls back to the `[x x]` case which is definitely not what we want.

If we have something in the following form: `<pre class="brush: ...">`, it's pretty safe to assume that things inside is highlightable (if not, I feel that the original code will break too anyway!), so it's safe to apply `xexpr->markdown`.

Potentially fix #165 (I'm not exactly sure if this is the issue that @LeifAnderson experienced or not), but this definitely fixes a bug that I am experiencing

### DIFF
--- a/frog/enhance-body.rkt
+++ b/frog/enhance-body.rkt
@@ -13,8 +13,7 @@
          "html.rkt"
          "params.rkt"
          "pygments.rkt"
-         "xexpr-map.rkt"
-         "xexpr2text.rkt")
+         "xexpr-map.rkt")
 
 (provide enhance-body)
 
@@ -29,12 +28,12 @@
 (define (syntax-highlight xs)
   (for/list ([x xs])
     (match x
-      [(or `(pre ([class ,brush]) (code () ,texts ...))
-           `(pre ([class ,brush]) ,texts ...))
+      [(or `(pre ([class ,brush]) (code () ,(? string? texts) ...))
+           `(pre ([class ,brush]) ,(? string? texts) ...))
        (match brush
          [(pregexp "\\s*brush:\\s*(.+?)\\s*$" (list _ lang))
           `(div ([class ,(str "brush: " lang)])
-                ,@(pygmentize (apply string-append (map xexpr->markdown texts)) lang))]
+                ,@(pygmentize (apply string-append texts) lang))]
          [_ `(pre ,@texts)])]
       [x x])))
 
--- a/frog/enhance-body.rkt
+++ b/frog/enhance-body.rkt
@@ -13,7 +13,8 @@
          "html.rkt"
          "params.rkt"
          "pygments.rkt"
-         "xexpr-map.rkt")
+         "xexpr-map.rkt"
+         "xexpr2text.rkt")
 
 (provide enhance-body)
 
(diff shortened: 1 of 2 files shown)


greghendershott:
Good catch.

I'm not 100% sure whether `xexpr->markdown` is the right thing to use. That's intended to make plain text "lite" markdown for humans to read (e.g. IIRC it's called by `xexprs->description` when we need to convert HTML to some plausible text for page metadata) and that will _not_ be a reserved character in HTML.

So for example `(xexpr->markdown 'gt)` is `""` -- not `">"` or `"&gt;"`.

Thinking ...

sorawee:
Please note what Brown PLT does for their websites:

https://github.com/brownplt/pyret.org/blob/master/src/page-template.html

sorawee:
Duh. I just realized that works only on non-post (i.e. page-template.html).

greghendershott:
Instead of `xexpr->markdown` should we maybe use [`xexpr->string`](https://docs.racket-lang.org/xml/index.html#%28def._%28%28lib._xml%2Fmain..rkt%29._xexpr-~3estring%29%29)?

frog/enhance-body.rkt
```diff
@@ -28,12 +29,12 @@
...
     (match x
-      [(or `(pre ([class ,brush]) (code () ,(? string? texts) ...))
-           `(pre ([class ,brush]) ,(? string? texts) ...))
+      [(or `(pre ([class ,brush]) (code () ,texts ...))
+           `(pre ([class ,brush]) ,texts ...))
        (match brush
          [(pregexp "\\s*brush:\\s*(.+?)\\s*$" (list _ lang))
           `(div ([class ,(str "brush: " lang)])
-                ,@(pygmentize (apply string-append texts) lang))]
+                ,@(pygmentize (apply string-append (map xexpr->markdown texts)) lang))]
```
greghendershott: To be clear, in my previous comment I meant should we use `xexpr->string` instead of `xexpr->markdown` _here_ -- in the argument to `map`?

Would that produce a valid string for Pygments?

sorawee:
That won't work:

```
> (require xml)
> (xexpr->string 'rsquo)
"&rsquo;"
```
but we want `'`.

sorawee:
@greghendershott should be correct now. Please review.

sorawee:
Note that Travis CI's failure is due to Racket server being down.

greghendershott:
I clicked `Restart build` on Travis and to me surprise it worked; looked like the Racket server is back.

I'm glad you thought about this longer and more clearly than me, and realized that `pygments-code` is the right place/level to fix this, and to do so using [`literal`](http://docs.racket-lang.org/scribble/base.html#%28def._%28%28lib._scribble%2Fbase..rkt%29._literal%29%29). Thanks!

PR #177, opened 2016-12-20
Add how to use Pygments in Scribble

sorawee:


sorawee:
@greghendershott I didn't see that you just did that lol

PR #180, opened 2017-02-27
Explicit use isPermalink="false" in Frog RSS feeds

gasche:
Thunderbird as a RSS client does not correctly render Frog-produced
blogs: the web link shown to the user is the guid flag, which is not
a URL (it is a mangled URL) and doesn't go anywhere when you click on
it.

The RSS currently generated looks like:

    <guid>urn:http-prl-ccs-neu-edu:-blog-2017-02-21-bullets-are-good-for-your-coq-proofs</guid>

According to the RSS specification, this is incorrect: there is
a `isPermaLink` attribute that indicate whether or not the guid can be
interpreted as a URL, but its default value is `"true"`. So if we
include a <guid> element in the output, we *must* specify
`isPermaLink="false"`. See the specification:

  https://validator.w3.org/feed/docs/rss2.html#ltguidgtSubelementOfLtitemgt

### DIFF
--- a/frog/feeds.rkt
+++ b/frog/feeds.rkt
@@ -119,7 +119,7 @@
     ()
     (title ,title)
     (link ,item-uri)
-    (guid () ,(str "urn:"
+    (guid ([isPermaLink "false"]) ,(str "urn:"
                    (our-encode (current-scheme/host))
                    ":"
                    (our-encode uri-path)))


greghendershott:
@gasche Thank you!

gasche:
(I forgot to credit @cpitclaudel for spotting the bug and finding the relevant part of the RSS specification. Writing the patch was actually trivial from there -- in particular thanks to the new CONTRIBUTING.md document.

greghendershott:
As long as we're giving thanks where it's due -- very big thanks to @tger for adding that `CONTRIBUTING.md`!

gasche:
Indeed; this document is particularly useful to me because I'm not a frequent Racket programmer, so whenever I want to do some changes in Frog I don't remember the rituals to try the modified version -- which I suppose remain in the working memory of people that actively contribute to other Racket programs.

PR #181, opened 2017-03-07
example/custom.css: include a useful snippet to avoid image overflow

gasche:
We needed this on the PRL blog, and I don't see any situation where you would not want this by default, so I think that upstreaming it would be useful.

### DIFF
--- a/example/css/custom.css
+++ b/example/css/custom.css
@@ -11,3 +11,16 @@ pre {
     white-space: pre;
     width: inherit;
 }
+
+article img {
+  /* Make sure that images within blog articles never take more width
+     than the article width; `height: auto` ensures the height is also
+     resized proportionally.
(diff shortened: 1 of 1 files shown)


greghendershott:
I know just enough CSS to be dangerous, so:

The `img` part seems harmless.

But I'm wondering what it means for `article` elements to max at 100%.

For example, just above in this file is some CSS that tries to let code blocks run past the right edge rather than wrap. How would these interact?

gasche:
`article img` means any `img` that is a transitive child of `article`, I believe, the disjunction would be `article, img`.

greghendershott:
You're right. I saw a `,` that wasn't there.

greghendershott:
Merged -- thanks!

PR #190, opened 2017-04-20
Added KaTeX widget, and MathJax compatibility layer

wilbowma:
Added a KaTeX widget. Unfortunately, KaTeX and MathJax have different APIs, and Markdown only knows about MathJax. If you want to use LaTeX inline in Markdown, then you'll need to use the MathJax compatibility layer. If you use inline HTML to use KaTeX directly, this isn't necessary.

### DIFF
--- a/frog/widgets.rkt
+++ b/frog/widgets.rkt
@@ -297,3 +297,43 @@
                 src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.0/MathJax.js?config=TeX-AMS-MML_HTMLorMML">
         </script>
         })
+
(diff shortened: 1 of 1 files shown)


wilbowma:
Also, MathJax and KaTeX aren't 100% compatible, so you may not always be able to use KaTeX. I also haven't been able to test this thoroughly.

wilbowma:
Maybe be worth holding off on this, as this compatibility layer is being improved and merged into KaTeX https://github.com/Khan/KaTeX/pull/680

greghendershott:
@wilbowma Ah -- nice!

avelino:
Great feature, we need to review the implementation (pull with master branch)

wilbowma:
Oh thanks for the reminder. Yeah, this compatibility later got merged upstream. I'll have to take another look at the patch and test it. Unfortunately, I've got little time until late November.

PR #192, opened 2017-04-21
Support for less JavaScript.

wilbowma:
I'm not sure this really belongs in Frog; it's simple enough to do manually, but, I want to share and encourage less JavaScript to exist on the internet.


wilbowma:
Oh this is a foolish idea; this won't work with the responsive design.

PR #193, opened 2017-04-21
Initial proposal for plugin API

tger:
This is a bare-bones implementation of a Frog plugin API. I am issuing this pull request in order to get feedback on the general approach.

I've drawn inspiration from how DrRacket plugins work. The idea is that the all one needs to do to install a plugin is a to install the plugin package (normally through the Racket package manager). Frog will find installed plugins automatically (through a magic `frog-plugin` symbol in the plugin `info.rkt`) and load them at startup.

Right now there are only two extension points (hooking into the `enhance-body` pipeline, and registering a cleanup handler). But already this is sufficient to convert the responsive image code to a plugin (modulo a way for plugins to read variables from `.frogrc`). The Pygments code would be another nice candidate (which would also do away with the Python and Pygments dependancy for a basic Frog installation).

I have pushed a "Hello World" plugin to https://github.com/tger/frog-hello.git.

Thoughts?


greghendershott:
Thanks!!

A quick reaction and brain-dump:

IIUC, there is a chain of "body enhancers", each of which is a function from `(listof xexpr?)` to `(listof xexpr?)`.  But also, they are allowed to have side-effects (such as making image files on disk), ergo the need to supply a `clean` function, too.

- Just for the sake of argument: I suppose an alternative would be to say, look, if they can have side-effects on the file system, anyway, why not cave in and just go all the way -- why not give them a `/path/to/files/` and they just party directly on HTML files on disk?  An advantage is they needn't even be written in Racket. A disadvantage is each will be re-parsing HTML.  I'm not saying I love this idea, but for the sake of completeness, it's a choice.

- I'm not 100% sure about the chain mechanics. Maybe the order could matter?  Maybe the user should be able to disable one without having to `raco pkg remove` it?  So maybe this can't work _quite_ like how DrRacket discovers plugins.

- > (modulo a way for plugins to read variables from .frogrc)

  Frog can `dynamic-require` the plugins, they can each `(require frog/params)`, and they can use the `current-xxx` parameters whose values will be set from `.frogrc`.

  I think a related point is we'd need to carefully prescribe what _else_ a plugin may do. If they `require` any frog module and use anything, it's not likely to end well.

- I could also see having an extensibility interface for source file formats (e.g. someone wants to add Org format). And of course dog-fooding it by reimplementing the hardwired/built-in markdown and scribble formats. Probably that's best to leave as a separate issue, but mentioning in case there's some overlap in plugin mechanics.

tger:
I would suspect most Frog users would prefer to write their plugins in Racket, and there is always the option of forking a subprocess from Racket. But having a generic plugin which directly implements this use-case would perhaps be a good idea for people preferring not to use Racket. 

>Frog can dynamic-require the plugins, they can each (require frog/params), and they can use the current-xxx parameters whose values will be set from .frogrc.

What I was referring to was for plugins to have a way to add their own parameters (but whose values are read from `.frogrc`). I've hacked up an implementation of this, though not so pretty at the moment since bindings created at runtime need to be referenced from `eval` forms.

>I think a related point is we'd need to carefully prescribe what else a plugin may do. If they require any frog module and use anything, it's not likely to end well.

Right, hence the API. A plugin is meant to require only `frog/plugin-api`.

tger:
I settled on exporting get-config in the plugin API. I think most plugins won't need parameters,  module top-level bindings probably suffice (initialized using `get-config`). And if they do need a parameter they can just create one in the plugin module. (I may to this in the responsive image plugin). See the updated Hello World-plugin for example.

I am in the process of refactoring the responsive image code into a plugin. I noticed that it makes use of some path-related functionality from the _frog/paths_ module, I'll see if the plugin API should export them or if I should go for some other solution.

tger:
Regarding disabling plugins without removing the package, I guess a Frog command line switch and/or config file setting would be the possible options.

tger:
How do you feel about placing the Pygments support in a plugin? It would be real easy.

greghendershott:
The proposed plugin API is making think a lot about:

- Who is Frog for?  Is it a blogging tool for "everyone" even non-programmers?  Or (realistically?) is it for programmers?

- Racketeers tend to embrace the idea that everything is a program (sometimes in a special `#lang`). For example slideshows are programs. Books are programs.  Of course dynamic web sites are programs. So, why isn't your static blog site a program?? In other words: Why is Frog so "un-Rackety"?

Years ago when I started Frog,  I figured it would be neat if all the kinds of people who used Jekyll or Octopress could instead use Frog. They'd need to install Racket instead of Ruby, is all.  And so it is very much a command-line tool that you use -- not a program that you write. But should it remain that way?

---

What if Frog were tweaked to be used as a library (possibly providing a `#lang` or two, if that's actually useful), and you write a simple program?  In the default/common case, it's probably a single not-very-long `.rkt` file.

In that scenario:

- There wouldn't need to be all the existing hoo-hah with reading  a `.frogrc` file. Instead you'd directly set some Racket parameters, and/or pass some different argument values to some functions, and/or call certain functions (or not).

- There wouldn't need to be "plugin packages"; instead there'd be packages. There wouldn't need to be a plugin API; instead you'd simply `require` and use the extra packages you want, in your blog-as-program.

---

This is really nagging at me because I feel like that's the right new direction to go.

Admittedly it raises other questions. Including, does Frog-as-tool remain available indefinitely (just deprecated)?

Also the right thing to do might be (a) proceed with this plugin API soon, and (b) do what I'm talking about, later.  I realize @tger you've been trying to add support for responsive images for a long time now!

greghendershott:
p.s. The difference might turn out to be smaller than it seems. Effectively it might mean that instead of a `.frogrc` you have a `blog.rkt`. On the one hand, that's a pretty big difference. On the other hand, `raco frog` might still exist as a command-line that "drives" functions in your `blog.rkt`. I'm just spitballing here, there are many ways to do this and I'm not saying I feel I know the best way(s), yet.

tfeb:
This is a response to @greghendershott's 'who is frog for' comment, from my perspective.

I don't know if frog is for programmers or non-programmers, but I would be happy if it was just for programmers (ie I am happy to have to write some small amount of code to create the blog).

I think it would be fine if a blog was a program in either Racket or some customised little language.  However:

- it is essential that this can be driven from the Unix command line, so I can write Unix commands around it which are driven by `make` & talk to git and so on (I don't currently create new entries this way but I can see that I might one day do so);
- it is essential that the source text is markdown (or that markdown & other common formats remain available) and that it is compatible with what exists now so I don't have to rewrite old entries;
- it is essential that the result is static;
- it is essential that any new thing either is compatible with existing configuration or provides a well-defined, once-off migration path which results in a readable and maintainable configutation.

In the above 'essential' means 'essential for me': it would be perfectly reasonable for you to decide that what I wanted did not matter (it's your tool after all) and I would not complain about that: I probably wouldn't use Frog though, and I would be sad about that.

---

In my case I came to Frog because I was looking for something that ate markdown, that was simple, visually attractive and entirely static, and that didn't require me to install & support some vast infrastructure full of security problems in a language in which I have no interest.  Well, I'm a Lisp person and I already played with Racket so Frog was ideal for the last point.  But it also was ideal for the other points: it was simple enough that when it didn't initially do quite what I wanted I could change it so it did (and get the changes accepted: thanks!) and the generated blog is also simple, can be visually attractive and is static of course.

Frog is a bit of software that makes me feel good about software, in a way that almost nothing does any more.

I'd be sad if Frog lost any of those properties, although, again, it's not my tool and if what you want from it is different then what I want, well, what you want is what counts.

greghendershott:
@tfeb Thank you for the feedback that's really helpful! I understand and agree.

I want Frog to remain simple and reliable to use (and maintain). At the same time, over the years people understandably want more configurability and capabilities. This is a pretty typical tension, in my experience (unless no one uses the software at all, in which case no problem :smile:).

Roughly what I have in mind is that Frog would deprecate `.frogrc` in favor of a user-defined `blog.rkt`. A default `blog.rkt` would be supplied in `example/` and copied for `raco frog --init`.

The `blog.rkt` would:

- Set configuration parameters. e.g. instead of `author = Jane Doe` you'd have `(current-author "Jane Doe")`.  It's likely some of these parameters wouldn't even be needed anymore, because next item.

- Supply code for things like `enhance-body`.  Probably there'd be some threading macro like this code literally from Frog today:

  ```
  (define (enhance-body xs)
    (~> xs
        syntax-highlight
        add-racket-doc-links
        auto-embed-tweets))
  ```

  And you could change this to (say) add responsive images @tger or KaTex @wilbowma or any other `(Xexprs -> Xexprs)` function. The functions you `require` from a module supplied by Frog, or from a "third party" package, or even define locally in that `blog.rkt` for your sole personal use.

- Maybe there is also a `read-post` function, which would let you add `(Path -> Xexprs)` functions that know how to read Org-mode format posts, or whatever, for different file extensions.

So the idea here mainly is that a `conf` style file like `.frogrc` was kind of a mistake. Rather than continue down that path for "plugins", let's step back and say -- hey people can write some simple Racket code to express whatever they want.  TL;DR: `(> rkt conf)`.

tfeb:
@greghendershott I think if it was explained (however roughly) how to change a `.frogrc` to the new config, this would meet my criteria for an easy, once-off migration, and once it was done a `blog.rkt` would be a much preferable format.  So this sounds like a straightforwardly good idea to me (sorry if my previous comment made me sound like I didn't think this: I just have had too many experiences of stuff I rely on getting broken).

I'd be happy to try out any test versions of this.

greghendershott:
@tfeb I'm glad you're skeptical. Fortunately, yes: I didn't mean it sound like, "_holds breath_ Whoa, everything is a program, man! _exhales_". :wink:

If we had thousands of enterprise customers then of course we could sell half-day webinars where people earn Certified Frog Migration credentials. But. I think some simple documentation should suffice.

- Mostly it would amount to "change instances of `key = val` to in `.frogrc` to `(current-key val)` in `blog.rkt`.

- In some cases the parameters would just disappear. We wouldn't need a `current-auto-embed-tweets` parameter corresponding to the `auto-embed-tweets` setting in `.frogrc`.  Not when, in a `blog.rkt`, you could just change the `enhance-body` function to remove the call to `auto-embed-tweets`. (That's the kind of ceremony/mechanics I'd like to remove, and, avoid creating more of.)

- People would be _free_ to do more elaborate things in `blog.rkt`, but they shouldn't need to simply to keep an existing blog working.

_Maybe_ I could supply a `blog.rkt` that isn't a model for new blogs, but simply for backward compat; it implements the status quo including reading `.frogrc` even the "silly" params.  Not sure. "Just-works" is good, unless it's so automagical that people get confused when they want to change an existing blog. Trade-offs.

tfeb:
@greghendershott I'm reasonably sure that the two (weird) tricks for having thousands of enterprise customers is:

- make whatever it is you are selling be unspeakably appalling to install and use;
- offer very expensive training courses on how to install & use it.

If you make it hard enough to install and use then it does not even need to work, since no-one will ever get that far, and you can simply become enormously rich selling courses.

tger:
There would still be a need to distribute larger chunks of supporting code (such as responsive images support, KaTeX..) for users who want them. Putting those in Racket packages is probably still a good idea (copy/paste from a wiki only scales so far..). The only difference would be that they would need to be manually `require`d in said `blog.rkt`. 

Whether going with the framework-style Frog (its current incarnation) or a library-style Frog is as @greghendershott already remarked probably decided by whether the target user identify as a Racket programmer or not. It seems like the fate of most software written in Racket ultimately end up on this path. It's a bit of a shame since it implies that people who are not interested in hacking Racket pretty much have no choice but to resort to software written in Python or Ruby despite the fact that the Racket alternative may be of higher quality.

greghendershott:
I pushed commit 5de5eef on a branch named `blog-as-program`. It's still just WIP, as described in the commit message.

greghendershott:
@tger 
> There would still be a need to distribute larger chunks of supporting code (such as responsive images support, KaTeX..) for users who want them. Putting those in Racket packages is probably still a good idea (copy/paste from a wiki only scales so far..). The only difference would be that they would need to be manually required in said blog.rkt.

Yes. Exactly. Although I'm taking a different direction on the plugin API per se (sorry!), the extension itself would be very similar to what you'd already sketched out (thanks!):  A Racket package that provides `enhance-body` and `cleanup` functions.

> Whether going with the framework-style Frog (its current incarnation) or a library-style Frog is as @greghendershott already remarked probably decided by whether the target user identify as a Racket programmer or not. It seems like the fate of most software written in Racket ultimately end up on this path. It's a bit of a shame since it implies that people who are not interested in hacking Racket pretty much have no choice but to resort to software written in Python or Ruby despite the fact that the Racket alternative may be of higher quality.

> For Racket programmers though, this is obivously a good thing and in the case of Frog it paves the way for Emacs-level customizability, something a plugin API could never achieve.

I think Emacs is a good analogy. A lot of people configure in Emacs Lisp just fine without being experts. I think (hope?) that most people's `blog.rkt` for the sorts of things you can do in Frog, today, won't be very complicated -- really no more so than their `.frogrc`. See the `example/blog.rkt` in the commit I just pushed. (And that's "noisier" than it will be, due to some temporary FIXME comments.)

Of course some people hate s-expressions so maybe the non-`conf` "smell" will be off-putting to some.

greghendershott:
Pushed another WIP commit on the [`blog-as-program` branch](https://github.com/greghendershott/frog/commits/blog-as-program).

tfeb:
@greghendershott:  The obvious answer is to have an alternative which uses markdown as the config file format.

This is almost serious.

tger:
I think this is shaping up pretty nicely.

However, I do not that providing _enhance-body_ and _clean_ procedures should be mandatory (in particular not the latter). How about arranging so that things are still OK if they're not exported from `blog.rkt`? (in which case they would default to the identity and empty function, respectively)

tger:
I portend a lot of more activity on the wiki once this goes out into the wild. Perhaps it will even become a fun way of getting into Racket-hacking for some people!

greghendershott:
@tger Thanks.

> However, I do not that providing enhance-body and clean procedures should be mandatory (in particular not the latter). How about arranging so that things are still OK if they're not exported from blog.rkt? (in which case they would default to the identity and empty function, respectively)

Good question.

My hunch is there will be less support -- which is relevant to my interests! -- if the functions are explicit and required.

(I understand the convenience of implicit/optional for something people need to do frequently. But most people won't create a `blog.rkt` by hand, ever: The initial/default `blog.rkt` is supplied automatically -- either `raco frog --init`, or, the first time you run this new version on an existing blog and it auto-converts `.frogrc` to the equivalent `blog.rkt`.)

tger:
> My hunch is there will be less support -- which is relevant to my interests! -- if the functions are explicit and required.

Well, as long as we do not assume that Frog users are ignorant. (or we will end up with ignorant users!)

Did you consider using `racket/class`?

As for _clean_, since I suspect most people maintain their blog using Git, doing away with it entirely to just rely on `git-clean` may be an attractive option (it's simple and really fast).

greghendershott:
Closing while doing some housekeeping. Instead, starting with commit 005833f, we went with the new approach using a `#lang frog/config` `frog.rkt`.

PR #194, opened 2017-04-26
.frogrc => blog.rkt

greghendershott:
This was motivated by #193, where there is some background discussion.

The basic idea is that instead of a `conf` style `.frogrc`, each blog project has a new, user-supplied [`blog.rkt`](https://github.com/greghendershott/frog/blob/blog-as-program/example/blog.rkt). This `provide`s a few functions that Frog uses:

- `init` is called by Frog early on startup. The user may change various parameters, directly, that they used to via the indirection of the `.frogrc` style file.  So instead of `author = Jane Doe` we have `(current-author "Jane Doe")`. However a half dozen of these parameters can disappear, because of the next item:

- `enhance-body` is a function from x-expressions to x-expressions. It's called to process the contents of post and non-post pages.  Now, the user can directly call various functions (or not), such as ones supplied by Frog: `syntax-highlight`, `auto-embed-tweets`, and `add-racket-doc-links`.  These used to be controlled by the now-defunct parameters.

  Better still, the user may call _any_ function here. It could be one they define locally in `blog.rkt`. Or it could be from a third-party package -- such as one to do responsive images, or KaTeX math, or whatever. There needn't be any "plugin API"; if the user wants to call something, they call it.

- `clean` is just in case some body-enhancer generates extra files that Frog doesn't know how to clean.

---

A default `blog.rkt` is created for users in two situations:

- `raco frog --init` to create a new project
- The first time they run this newer version of Frog: It automatically creates the `blog.rkt` equivalent of their `.frogrc`; thereafter the latter is ignored

---

On the one hand, this isn't a huge change. There is no `#lang blog`. This isn't your-blog-is-a-program in the sense of your slideshow or your book. It's not full-on Rackety.

On the other hand, it's _more_ Rackety than before. And:

- It increases extensibility while avoiding ceremony/mechanics for "plugins".

- Anyway, the ability to keep using commands like `raco frog` is something users like @tfeb want.  It seems like a reasonable amount of change.

- The auto-created `blog.rkt` should make this smoothly compatible (fingers crossed).

- Although the `blog.rkt` won't be _quite_ as simple as a `conf` style file, it's not very much more complicated.

/cc @tger @wilbowma 

Comments welcome from all.


frog/private/frog.rkt
```diff
@@ -23,143 +26,124 @@
...
+      (""
+       "Delete generated files.")
+      (clean)]
+     #:once-each
+     [("-w" "--watch")
+      (""
+       "(Experimental: Only rebuilds some files.)"
+       "Supply this flag before -s/--serve or -p/--preview."
+       "Watch for changed files, and generate again."
+       "(You'll need to refresh the browser yourself.")
```
bennn: yourself.)

frog/private/frog.rkt
```diff
@@ -23,143 +26,124 @@
...
+      (""
+       "The port number for -s/--serve or -p/--preview."
+       "Supply this flag before one of those flags."
+       "Default: 3000.")
+      (set! port (string->number number))]
+     [("--root") path
+      (""
+       "The root directory for -s/--serve or -p/--preview."
+       "Supply this flag before one of those flags."
+       "If .frogrc has uri-prefix = /path/to/site/blog, try --root /path/to/site"
```
bennn: `.frogrc` is dead!

frog/private/upgrade-from-frogrc.rkt
```diff
@@ -0,0 +1,76 @@
...
+    (flush-output)
+    (with-output-to-file #:mode 'text #:exists 'error
+      blog.rkt (λ () (default frogrc)))))
+
+(define (default frogrc)
+  (define (get k d) (get-config k d frogrc))
+  (define (pw v)
+    (pretty-write v)
+    (newline))
+  (displayln "#lang racket/base\n")
```
bennn: I don't think "\n" is platform-independent.

frog/frog.scrbl
```diff
@@ -234,157 +230,45 @@ For examples of @tt{pygments.css} code highlighting styles see
...
+package. You may even define a function locally in @tt{blog.rkt}.}
 
-@defcfg[pygments-linenos? boolean? true]{Whether Pygments should
-include line numbers in its HTML output.}
+@defproc[#:link-target? #f
+         (clean) any]{
 
-@defcfg[pygments-cssclass string? "source"]{The CSS class Pygments
-should use for the wrapping @tt{<div>} tag.}
+Called during @tt{raco frog @literal{--}clean}.}
```
greghendershott: Oh wow. That's awesome. How did I overlook these? Thanks!

frog/frog.scrbl
```diff
@@ -984,3 +868,152 @@ In your Frog project directory, create an output directory named
...
+How many items to include in feeds? Older items in excess of this will
+not appear in the feed at all.}
+
+@defparam[current-decorate-feed-uris? v boolean? #:value #t]{Decorate
+feed URIs with Google Analytics query parameters like
+@tt{utm_source}?}
+
+@defparam[current-feed-image-bugs? v boolean? #:value #t]{Insert in
+each feed item an image bug whose URI is decorated with Google
+Analytics query parameters like @tt{utm_source}?}
```
greghendershott: Although I like to document a boolean as the question it answers, it's good to know it can be confusing. I'll rephrase.

frog/frog.scrbl
```diff
@@ -984,3 +868,152 @@ In your Frog project directory, create an output directory named
...
+located.}
+
+
+@section[#:tag "body-enhancers"]{Body enhancers}
+
+@defmodule[frog/enhance-body]
+
+@defproc[(syntax-highlight
+          [xs (listof xexpr/c)]
+          [#:python-executable python-executable string?]
```
greghendershott: If I understand your question correctly: yes. For most people `"python"` should just work, and, that's used in the default `blog.rkt`.

frog/private/frog.rkt
greghendershott: I don't understand what you mean here?

frog/private/frog.rkt
greghendershott: 👍

frog/private/frog.rkt
bennn: The open parenthesis that starts before `You'll` is never closed.

frog/frog.scrbl
bennn: Okay. Maybe leave a note that `python-executable` gets passed to `find-executable-path`?

Reading the docs, I wasn't sure if `python-executable` needed to be the name of something in `/usr/bin`, or an absolute path, or if it could even be an alias that I set in my `.bashrc`.

mbutterick:
Suggest `frog.rkt` instead of `blog.rkt`.

jaor:
On Thu, Apr 27 2017, Matthew Butterick wrote:

> Suggest frog.rkt instead of blog.rkt.

+1

greghendershott:
@mbutterick @jaor Hmm.

From _my_ perspective, I'm drowning in files and dirs named `frog` -- think of the old Gary Larson cartoon, "blah blah blah Ginger blah blah blah" -- so picking some _other_ name seemed awesome. To me.

But you're right. From a user's point of view, looking among the files in their project, `frog.rkt` is going to afford, "I'm the special file that Frog cares about."  Thanks for helping me see that!

Having said that, I just pushed a commit I'd been preparing, to change how the-file-soon-to-be-formerly-known-as-`blog.rkt`is prepared. I'll do a separate pass and commit to rename.

greghendershott:
> Suggest frog.rkt instead of blog.rkt.

@mbutterick @jaor done in commit fa6b690

jaor:
thanks greg, and matthew who actually had the idea in the first place! :)

example/frog.rkt
```diff
@@ -0,0 +1,42 @@
...
+  (current-scheme/host "http://www.example.com")
+  (current-title "My Blog")
+  (current-author "The Unknown Author"))
+
+;; Called once per post and non-post page, on the contents.
+(define/contract (enhance-body xs)
+  (-> (listof xexpr/c) (listof xexpr/c))
+  ;; Here we pass the xexprs through a series of functions.
+  (~> xs
+      (syntax-highlight #:python-executable "python"
```
tger: Any particular reason not to rely on `(find-executable-path "python")`?

greghendershott: @tger This flexibility was added in PR #125 for someone juggling Pythons 2 and 3.

greghendershott:
@tfeb You mentioned you'd like to try this. I think it has settled down enough, now.

I dog-food-ed it on my own blog, today:

1. `raco frog`. It made a new `frog.rkt` from my `.frogrc`. Compared; looked good.
2. `raco frog --clean --build` to make it regenerate everything.
3. `git status` said no output files had changed as a result.

(You could skip directly to step 2, which would also do 1. I listed it this way just to emphasize the one-time auto-creation of `frog.rkt`.)

example/frog.rkt
tger: I think resorting to `virtualenv` or such would be nicer to not let Frog worry about this but no big deal I guess.

tger: How about making "python" the keyword default value?

LeifAndersen:
Would you be opposed to it being `#lang blog`. Or rather, something like `#lang info`, so that you don't actually need to type out the `provide` line?

I'm fine one way or the other, as well as making the `#lang`. This is more a question of design.

Either way, I love this direction, and thank you.

greghendershott:
I'm not opposed to a `#lang`. It could save people needing to do an explicit `provide`, as you say. Also avoid some of the explicit `require`s. Although that's not a huge advantage, it's something.

Also, I suppose it would nudge people away from using `#lang racket` instead of `#lang racket/base` (and in the config file casually undermining frog carefully using the latter).

I wouldn't want it to be restricted like `#lang info`, which has no `#%top-interaction`, and, you can't define functions. If someone wants to define a little body-enhancer locally in their `frog.rkt`, they should be able to.

So I'll take a look at doing this.

p.s. I'd feel presumptuous calling it `#lang blog` (_the_ lang for blogs!) :smile:. Also it's not really in the same league as say `#lang slideshow`.  So probably `#lang frog`. Or even more honestly, `#lang frog-config`.

bennn:
+1 for `#lang frog/config`, I think you should be able to make a language-as-dotfiles `config.rkt` pretty easily

http://blog.racket-lang.org/2017/03/languages-as-dotfiles.html

greghendershott:
@benn 
> what would you do to automatically insert the `provide`s? Override `#%module-begin`?

That's what I tried last night. I found that it doesn't work unless the `provide` uses the original syntax for the identifiers. For instance, the following seems to work. Is there a more elegant, concise, and/or correct way to do this?

```racket
(begin-for-syntax
  (define (syntax-e=? a b)
    (equal? (syntax-e a) (syntax-e b)))

  (define (fn-id mod-forms sym)
    (define (eq-sym? stx)
      (and (eq? sym (syntax-e stx)) stx))
    (for/or ([stx (in-list (syntax->list mod-forms))])
      (syntax-case* stx
          (define define/contract)
          syntax-e=?
        [(define          (s . _) . _) (eq-sym? #'s) #'s]
        [(define/contract (s . _) . _) (eq-sym? #'s) #'s]
        [_ #f]))))

(define-syntax (our-#%module-begin stx)
  (syntax-parse stx
    [(_ forms ...)
     ;; We need to provide the original identifier syntax.
     (with-syntax ([init         (fn-id #'(forms ...) 'init)]
                   [enhance-body (fn-id #'(forms ...) 'enhance-body)]
                   [clean        (fn-id #'(forms ...) 'clean)])
       #'(#%module-begin
          forms ...
          (provide init
                   enhance-body
                   clean)))]))
```

LeifAndersen:
I mean, if you just have your module-begin insert the identifiers to provide without any marks, that should work and is (imo) the right thing. I'll make that and post it here.

LeifAndersen:
Something like this:

```
#lang scratch

(define-syntax (~module-begin stx)
  (syntax-parse stx
    [(_ body ... )
     (with-syntax ([init (format-id stx "init")]
                   [enhance-body (format-id stx "enhance-body")]
                   [clean (format-id stx "clean")])
       #'(#%module-begin
          body ...
          (provide init
                   enhance-body
                   clean)))]))

(provide (except-out (all-from-out scratch) #%module-begin)
         (rename-out [~module-begin #%module-begin]))

(module* test (submod "..")
  (define init #f)
  (define enhance-body #f)
  (define clean #f))
```

Other things I would do:
1. Remove `provide` and possibly `#%provide`. I mean, they can still get it via a `require` if desired, but it will make things cleaner and error messages better.
2. Ya, I like the idea of calling this `#lang frog/config`. (`frog-config` doesn't quite mesh with the other language names quite so well. But I'm easy either way.)
3. In the actual macro, change the error message from 'provided identifier not defined', to something a little cleaner, like just <identifier> not defined. You can probably do this by catching the error at macro expansion time and re-raising it as the nicer name.
4. Umm...profit? thank you for doing this btw.

LeifAndersen:
Breaking hygiene here is also (imo) better. Because if the user has a macro that defines any of those three identifiers, the solution you proposed might accidentally capture that definition, which would be a very confusing bug.

greghendershott:
@LeifAndersen Oh -- yes, using `format-id` is simpler. Thanks!! I could even boil down:

```racket
(syntax-parse stx
  [(_ forms ...)
   (with-syntax ([(provides ...) (map (curry format-id stx)
                                      '("init" "enhance-body" "clean"))])
     #'(#%module-begin
        forms ...
        (provide provides ...)))])
```

---

> In the actual macro, change the error message from 'provided identifier not defined', to something a little cleaner, like just not defined. You can probably do this by catching the error at macro expansion time and re-raising it as the nicer name.

I was going to put a better error message in the more complicated code that we just nuked. Although I'm not yet sure how/where to "catch the error at expansion time", I'll think on that and see if I can figure that out.

LeifAndersen:
Oh, for that you probably want to use `local-expand`. Let me see if I can cook you up another example. ;)

greghendershott:
@LeifAndersen I'm happy because `local-expand` is what I guessed. :)

What I have at the moment:

```racket
(define-syntax (our-#%module-begin stx)
  (syntax-parse stx
    [(_ forms ...)
     (with-syntax ([(provides ...) (map (curry format-id stx)
                                        '("init" "enhance-body" "clean"))])
       (for ([needed-id (in-list (syntax->list #'(provides ...)))])
         (unless (for/or ([form (in-list (syntax->list #'(forms ...)))])
                   (syntax-parse (local-expand form 'module #f)
                     [(define-values (id) . _) (equal? #'id needed-id)]
                     [_ #f]))
           (raise-syntax-error 'frog/config
                               (format "Your frog.rkt must define a function named `~a'"
                                       (syntax-e needed-id))
                               stx)))
       #'(#%module-begin
          forms ...
          (provide provides ...)))]))
```

LeifAndersen:
Okay, here is a simple example of how you can do this:
```
(define-syntax (~module-begin stx)
  (syntax-parse stx
    [(_ forms ...)
     (with-syntax ([(provides ...) (map (curry format-id stx)
                                        '("init" "enhance-body" "clean"))])
       (with-handlers ([exn:fail:syntax?
                        (λ (e)
                          ;(displayln e)
                          (raise-syntax-error 'frog/config
                                              "required identifiers not defined"
                                              (exn:fail:syntax-exprs e)))])
         (local-expand
          #'(#%module-begin
             forms ...
             (provide provides ...))
          'module-begin
          '())))]))
```

One important thing to note: `exn:fail:syntax?` is **NOT** a good enough predicate for `with-handlers`, because this will catch _all_ syntax errors, which is not quite what you want.

A better predicate would be something like:

```
(lambda (x)
  (and (exn:fail:syntax? x)
       (string=? (exn-message x) "module: provided identifier not defined or imported for phase 0")))
```

Although you might want to just check to see if it has the text `not defined or imported`.

Anyway, hope that helps.

LeifAndersen:
@greghendershott Woops, just saw your second reply....reading it now.

LeifAndersen:
```
       (for ([needed-id (in-list (syntax->list #'(provides ...)))])
         (unless (for/or ([form (in-list (syntax->list #'(forms ...)))])
                   (syntax-parse (local-expand form 'module #f)
                     [(define-values (id) . _) (equal? #'id needed-id)]
                     [_ #f]))
           (raise-syntax-error 'frog/config
                               (format "Your frog.rkt must define a function named `~a'"
                                       (syntax-e needed-id))
                               stx)))
```

He, he, you really like manually traversing the whole module, don'tcha. ;)

Anyway, at a first glans this also looks fine, but I see two possible problems:

1. Because you are local expanding expressions one at a time, I'm not sure how it will act when a definition uses a macro defined in the previous expression. (It might be fine, I'm not sure.)
2. It also looks like you'll be expanding the module twice, which seems slow to me.

greghendershott:
@LeifAndersen Great points!  Here's what I have now, which uses your `(with-handlers .... (local-expand ....))` approach -- but tries to supply specific error messages.

```racket
(define-syntax (our-#%module-begin stx)
  (syntax-parse stx
    [(_ forms ...)
     (define provide-syms '(init enhance-body clean))
     (with-syntax ([(provides ...) (map (curry format-id stx "~a") provide-syms)])
       ;; Simply? #'(#%module-begin forms ... (provide provides ...))
       ;;
       ;; But if the user has failed to define one of the mandatory functions
       ;; we want to supply a helpful error message that specifically
       ;; identifies it. And of course preserve stx srcloc so their editor of
       ;; choice can open their frog.rkt.
       (define (fail-sym exn)
         (match (exn:fail:syntax-exprs exn)
           [(cons stx _) (syntax-e stx)]
           [_            #f]))
       (with-handlers
         ([(λ (e)
             (and (exn:fail:syntax? e)
                  (memq (fail-sym e) provide-syms)))
           (λ (e)
             (raise-syntax-error
              'frog/config
              (format "You must define a function named \"~a\"" (fail-sym e))
              stx))])
         (local-expand (syntax/loc #'(forms ...)
                         (#%module-begin
                          forms ...
                          (provide provides ...)))
                       'module-begin '())))]))
```

---

For example if `/tmp/foo.rkt` omits an `enhance-body` definition, the error is:

```
; /tmp/foo.rkt:1:0: frog/config: You must define a function named "enhance-body"
;   in: (#%module-begin (define (init) (void)) (define (-enhance-body xs) xs) (define (clean) (void)))
```

LeifAndersen:
1. Do you just want to print out the first missing identifier, or all of them. If you only want to print out the first one this works fine.

2. This looks like it will work for missing identifiers, what about for identifiers defined twice?

greghendershott:
1. I'm fine with error-ing one at a time, just as with the "raw" error does.

2. This doesn't inadvertently catch duplicate identifiers errors, they're displayed as usual.

greghendershott:
Re 2: But your point is correct. Since I posted the code here, I'd gone ahead and added a `(regexp-match? #rx"provided identifier not defined or imported" (exn-message e)))` condition, like your original example.

LeifAndersen:
Okay cool. In that case:

LGTM

Thank you again. 

:D

example/frog.rkt
greghendershott: > How about making "python" the keyword default value?

OK. Done in https://github.com/greghendershott/frog/pull/194/commits/27a3af4fd3c33834ff97f2a145980f9d38981ae9#diff-758af76c72ceeace13c505f9d6419389R25

greghendershott:
Closed via commit 005833f -- a rebase/squash of this PR's [`blog-as-program` branch](https://github.com/greghendershott/frog/tree/blog-as-program) (which I'll keep around for awhile for reference.)

Thanks everyone for the feedback!

tfeb:
@greghendershott: I realise I'm late to the party, but I have now tried this.  It's fine: it all just worked for me, including my rerooted blog and other slightly odd things.  So, thanks.

PR #200, opened 2017-09-01
Nicer error message when date is malformed

### LINKED ISSUE #199: Bad Date gives a Contract Error
If you make a blog post with an invalid date like `2017-04-38T00:00:00` you get an unhelpful contract violation:

```
date: contract violation
  expected: (integer-in 1 31)
  given: 38
  context...:
   write-post-page
   /home/max/racket/share/racket/collects/racket/contract/private/arrow-higher-order.rkt:340:33
   /home/max/.racket/6.6/pkgs/frog/frog/frog.rkt:246:2: for-loop
   /home/max/.racket/6.6/pkgs/frog/frog/frog.rkt:173:0: build
   /home/max/racket/share/racket/collects/racket/cmdline.rkt:450:10: call-handler
   (submod /home/max/.racket/6.6/pkgs/frog/frog/frog.rkt main): [running body]
   /home/max/racket/share/racket/collects/raco/raco.rkt: [running body]
make: *** [Makefile:7: blog] Error 1
```

As a user this looked like an internal error in frog to me, because I expect contract errors to mean that internal invariants are broken (i.e., a bug).
A better error would be to report that there was an error in parsing/validating the blog post specifically naming the post that was problematic (the biggest problem with the above error).

I'd be happy to fix it myself, seems like you could either just catch this contract error and add the extra context that you were parsing the blog post, or better yet to make this a parse error and direct it to the line where it occurred.

maxsnew:
Fixes #199 .
Basically the fix you suggested, but using `exn:fail:contract?` instead of `exn:fail?` to avoid potentially hiding other bugs.

Tested it manually by changing one of the examples to have a bad date.

### DIFF
--- a/frog/private/posts.rkt
+++ b/frog/private/posts.rkt
@@ -129,8 +129,15 @@
             #:when (member k '("Title" "Date" "Tags" "Authors"))
             (hash-set h (string-trim k) (string-trim v))]
            [s (warn s) h])))
+     (define d
+       (with-handlers ([exn:fail:contract?
+                        (λ _
+                          (err "`Date` is malformed, must be yyyy-mm-ddThr:mn:sc format"))])
+         (define d0 (hash-ref h "Date" (λ _ (err "missing `Date`"))))
+         (date->date-struct d0)
+         d0))
      (list (hash-ref h "Title" (λ _ (err "missing `Title`")))
-           (hash-ref h "Date" (λ _ (err "missing `Date`")))
+           d
            (append (~>> (hash-ref h "Tags" "")
                         tag-string->tags)
                    (~>> (hash-ref h "Authors" "")
--- a/frog/private/bodies-page.rkt
+++ b/frog/private/bodies-page.rkt
@@ -159,6 +159,7 @@
                [xs xs]))
   (string-join (map xexpr->string xs) ""))
 
+;; Raises a contract error on invalid date
 (define (date->date-struct YYYY-MM-DD-string)
   (match YYYY-MM-DD-string
     [(pregexp "(\\d{4})-(\\d{2})-(\\d{2})" (list _ y m d))


frog/private/posts.rkt
```diff
@@ -129,8 +129,15 @@
             #:when (member k '("Title" "Date" "Tags" "Authors"))
             (hash-set h (string-trim k) (string-trim v))]
            [s (warn s) h])))
+     (define d
+       (with-handlers ([exn:fail:contract?
```
greghendershott: This is too specific and I think you should use `exn:fail?`. For example, when `date->date-struct` is given a bad format string, then its `match` will fail . `match` fails with `exn:fail?` so you won't catch that here.

(If as you mentioned you're concerned about hiding other bugs: Then I think the way to address that is to wrap only the call to `date->date-struct` in the `with-handlers`?)

greghendershott:
About the Travis CI failures running the tests:

1. Run `make test` locally and you'll see them.

2. See my inline comment about catching `exn:fail?` instead of `exn:fail:contract?`.

3. There are some existing tests, in the `(module+ test __)` form right below the code you're changing. They were designed only to test the splitting/parsing of the meta-data. As a result the tests supply only dummy date strings like `"date"`. Obviously those aren't valid date strings, so `date->date-struct` is barfing on them. Now that you're requiring valid date strings early on, here, I think you'll need to update each such test with some example date string that's _valid_.  I think then the tests will pass when you `make test` locally, and you'll be all green on Travis CI.

Thank you again for offering to help!

greghendershott:
p.s. 4. If you _really_ wanted to be awesome, you could add a new test, using [`check-exn`](https://docs.racket-lang.org/rackunit/api.html#%28def._%28%28lib._rackunit%2Fmain..rkt%29._check-exn%29%29), with an intentionally bad date.

(If you haven't used `check-exn` before, one gotcha is to remember to enclose the test expression in a thunk -- `(lambda ()  ___ your test ___)`.

maxsnew:
Just as a heads up, I'm at ICFP/other workshops this week, but I'll act on your suggestions when I get back.

PR #203, opened 2017-10-07
Allow 'raco frog -h' to work outside of a project

belph:
When running `raco frog -h` outside of a project, one of two thing will happen:
- At best, a new `frog.rkt` file will be generated in your current directory based on your `~/.frogrc` file.
- At worst, a new `frog.rkt` file will be produced and an error message will appear because `~/.frogrc` doesn't exist.

This PR tweaks `private/main.rkt` to only run side-effecting operations and configuration loading when `-h` or `--help` is not passed.

### DIFF
--- a/frog/private/main.rkt
+++ b/frog/private/main.rkt
@@ -40,25 +40,28 @@
 
 (define (main)
   (printf "Frog ~a\n" (frog-version))
+  (define help-passed? (ormap (curryr vector-member (current-command-line-arguments)) '("-h" "--help")))
   (parameterize ([top (find-frog-root)])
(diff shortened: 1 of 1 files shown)


greghendershott:
Thank you very much for submitting this! I've been swamped and only had a chance to look at it carefully today. Partly I was nudged by #204.

What you're doing here is really good! I think I might try to handle a few more cases with my own commit.

greghendershott:
What I have so far: https://github.com/greghendershott/frog/commits/issue-204

greghendershott:
I ended up merging some other commits to address this. But thank you very, very much for submitting this!

PR #207, opened 2018-02-09
(wip) Editing older/newer-links function for Bootstrap 4

sleibrock:
So after doing the Bootstrap 4 transition, I realized that the paging between newer and older posts was not properly adjusted for Boostrap 4 (my bad). Here I adjust the function `older/newer-links` underneath `widgets.rkt`. I implemented the same code on my website, so I figured I'd PR it to Frog as well.

I'm not certain if I'm doing something wrong or not, but I couldn't quite get it to work under a blind test, ie: on a fresh project, generating pages would not use this change for some reason (not certain if it's because it's using my local Frog install or not). Is there anything I'm missing?

### DIFF
--- a/example/_src/post-template.html
+++ b/example/_src/post-template.html
@@ -9,6 +9,33 @@
     @twitter-share-button[full-uri]
     @google-plus-share-button[full-uri]
     @disqus-comments["shortname"]
-    @older/newer-links[older-uri older-title newer-uri newer-title]
+    @;(older/newer-links[older-uri older-title newer-uri newer-title])
+
+    @(define (page-navigation prev-uri prev-title new-uri new-title)
+      @list{
+        <div class="row justify-content-center">
+          <nav aria-label="Page Navigation">
+	    <ul class="pagination">
+	      @(when prev-uri
+	        @list{
+	          <li class="page-item">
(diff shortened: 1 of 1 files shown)


greghendershott:
We can't change `older/newer-links`, because that would break older blogs still using Bootstrap 3: If they get a newer version of Frog, then the next time they regenerate, their `post-template.html` would call the newer `older/newer-links` and get back incompatible Bootstrap 4 HTML.

Fortunately, I think you can simply inline `older/newer-links` into the example `post-template.html` -- copy its `@list{ ___ }` expression into `page-template.html`, then modify to be compatible with Bootstrap 4.

sleibrock:
I can inline it into `post-template.html` if you want then, I can just copy what I have from my own blog.

greghendershott:
@sleibrock Will you still be able to do this, or should I try to do it instead?

sleibrock:
I can do it, I was just a little preoccupied for the last few days.

greghendershott:
Merged -- thanks!

(Why the delay: GitHub didn't send me an email about the commit (I think it only does so for comments?) so I didn't notice this, and then I was mostly offline for a week.)

PR #223, opened 2018-07-29
Be more flexible about MathJax version

tfeb:
This PR slightly enhances the `math-jax` widget:it now lets you specify the source URL and the
configuration file as keyword arguments (`src` and `config`) with reasonable defaults.

So for instance

        @math-jax[#:src "https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.5/latest.js"]

is the right incantation to use whatever the latest MathJax is at the current time (see [documentation](http://docs.mathjax.org/en/latest/start.html)).

The default MathJax version version the widget uses is also now 2.7.5, up from 2.7.0

### DIFF
--- a/frog/widgets.rkt
+++ b/frog/widgets.rkt
@@ -267,16 +267,22 @@
         <script src="//gist.github.com/@|username|/@|id|.js"></script>
        })
 
-(define/doc (math-jax list?)
+(define/doc (math-jax [#:src src string?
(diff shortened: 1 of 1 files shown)


tfeb:
I don't understand why the Travis build failed: it looks to have timed out?  There seems to be no useful output anyway.

greghendershott:
I restarted the Racket 6.2 build. Those older Racket versions fail intermittently with a sqlite error during raco pkg install.

greghendershott:
Thanks!

PR #233, opened 2018-11-06
Improve --watch

sorawee:
1. There are several reasons why `frog.rkt` might not be loadable.
   We only want to show a nice warning that users should initialize the project
   when the exception is filesystem-related. Other errors like syntax errors /
   unbound ids should not be caught.

2. Improve `--watch` flag by allowing users to customize whether or not
   they want to trigger a rebuild or not. The default behavior is
   backward-compatible with the current behavior.

3. Don't print `\n` when beeping in `--watch`.

4. If an error happens during a build with `--watch`, prevent Frog from catastrophically fail. Note that this is especially common when editing a Scribble post file.

### DIFF
--- a/frog/config/private/load.rkt
+++ b/frog/config/private/load.rkt
@@ -20,7 +20,7 @@
 
         (define (load top)
           (define frog.rkt (build-path top "frog.rkt"))
-          (let ([fn (with-handlers ([exn:fail? cannot-find-frog.rkt])
+          (let ([fn (with-handlers ([exn:fail:filesystem? cannot-find-frog.rkt])
                       (dynamic-require frog.rkt 'id))])
             (when fn (set! id fn))) ...)
         (provide load))))
--- a/frog/frog.scrbl
+++ b/frog/frog.scrbl
@@ -924,6 +924,9 @@ This may be an absolute or relative path. If relative, it's relative
 to the project top directory, i.e. to where @secref["config"] is
 located.}
 
+@defparam[current-watch-rate rate exact-positive-integer? #:value 5]{
+The number of seconds to wait between each change watching.}
+
 @defparam[current-rebuild?
           rebuild?
           (path? (or/c 'create 'delete 'modify) . -> . boolean?)]{
@@ -934,8 +937,9 @@ on every file change (create, delete, modify) in the project. If @racket[rebuild
 returns @racket[#f], then Frog will not rebuild the project
 (for this particular change). Otherwise, Frog will rebuild the project.
 
-By default, any file change will trigger a rebuild, except the case where
-the changed file has an extension either @tt{html}, @tt{txt}, or @tt{xml},
+By default, @racket[rebuild?] will return @racket[#t] and also make a beep,
+except the case where the changed file has an extension either
+@tt{html}, @tt{txt}, or @tt{xml} which will return @racket[#f],
 as it is likely that the file will be an output.}
 
 @section[#:tag "body-enhancers"]{Body enhancers}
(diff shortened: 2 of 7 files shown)


sorawee:
Oh no, the build fails because I use a function that was added in 6.5.0.3. Do we really need to support Racket from 2 years ago? If the answer is yes, I will update the patch to avoid using the function.

frog/config/private/load.rkt
```diff
@@ -20,7 +20,7 @@
 
         (define (load top)
           (define frog.rkt (build-path top "frog.rkt"))
-          (let ([fn (with-handlers ([exn:fail? cannot-find-frog.rkt])
+          (let ([fn (with-handlers ([exn:fail:filesystem? cannot-find-frog.rkt])
```
greghendershott: That seems good -- thanks!

frog/frog.scrbl
```diff
@@ -924,6 +924,19 @@ This may be an absolute or relative path. If relative, it's relative
 to the project top directory, i.e. to where @secref["config"] is
 located.}
 
+@defparam[current-rebuild?
```
greghendershott: Although I don't necessarily object to this change, I don't understand the use case? Typically I use `--watch` because I want things to rebuild. Wouldn't I simply not use `--watch` if/when I don't want that?

frog/private/main.rkt
```diff
@@ -390,8 +391,9 @@
 
 (define (watch-callback path change-type)
   (when ((current-rebuild?) path change-type)
-    (build)
-    (displayln #"\007"))) ; beep (hopefully)
+    (with-handlers ([exn:fail? (compose1 displayln exn->string)])
```
greghendershott: This seems reasonable; thanks.

frog/frog.scrbl
sorawee: Actually it's the other way around: I want some `html` change to trigger a rebuild. Currently, it won't because all `html`s are blacklisted.

sorawee:
@greghendershott should be all right now.

frog/frog.scrbl
sorawee: A more radical question is, why having a blacklist at all? The watch is done only in the `_src` directory, but as I understand, the output html/xml/txt files would be outside of `_src`.

sorawee: So, if you think that there should be a blacklist, this patch simply gives you a finer control of that list, and you can merge it if everything looks good. On the other hand, if you think blacklist is not a good idea anymore, let me know and I can submit a change to remove all blacklist mechanism away.

sorawee:
Hello @greghendershott, do you get any chance to review this? Let me know if there's any problem with it!

greghendershott:
You happened to do a number of issues and PRs, at a time when I'm really swamped with work programming. Also, I haven't really been focused on Frog beyond simple maintenance, lately; it takes awhile to reload my L1 brain cache.

So, I've been slow to review. I have started to look at things a few times, but got deflected.

Although I'm not yet sure about the others, I think this PR seems good -- both the intended change and the implementation. Thank you for making the changes I suggested.

I will try to give it one last review later today, then probably merge it.

Also, if I notice anything trivial (e.g. a typo or phrasing in documentation), I'll just go ahead and change that myself in new commit, to avoid more cycles on this.

(It may take me awhile to get to the other PRs.)

greghendershott:
I squashed this down to 2 commits (the first 5, and the 6th), and rebased onto `master`. Tests passed again. Merged to `master` as commit 219301a and commit 1d6d42b.

Thank you.

sorawee:
Thanks! :)

PR #234, opened 2018-11-07
Make Disqus configurable

sorawee:
Allow to customize `identifier`, `title`, `url`, `category_id`. This patch is backward compatible with the current behavior.


greghendershott:
Merged; thanks!

Issue #214, opened 2018-04-17
Feature Idea: Arbitrary Post meta-data


LeifAndersen:
Right now a post can have metadata for `Title`, `Date`, `Tags`, and `Authors`. This is sufficient for most cases, but sometimes you want additional information in each post.

For example, lets say I wanted to enable comments, but I didn't like using disqus. I could follow this tutorial (http://hydroecology.net/using-github-to-host-blog-comments/), to use github issues to host comments for me. Unfortunately, there isn't a direct mapping from a github issue to the post itself, and so would need to be denoted for every post.

Right now, there is no good way to do that when using markdown, meaning I'd have to write all of my posts with scribble. I mean, you 'could' make a tag for the issue number, and parse that out, but that seems a bit...clunky. Instead, it might make sense to have extra meta-data be available to the `post-template.html` and `page-template.html` files, so that they can be used in formatting the page.

LeifAndersen:
The bit of code is here:

```
     (define h
       (for/fold ([h (hash)])
                 ([s (string-split plain-text "\n")])
         (match s
           [(pregexp "^ *(.+?): *(.*?) *$" (list _ k v))
            #:when (member k '("Title" "Date" "Tags" "Authors"))
            (hash-set h k v)]
           [s (warn s) h])))
```
in `posts.rkt`. It seems like it would theoretically be as easy as just adding all meta-data info to the hash table, rather than just `Title`, Date`, etc.

greghendershott:
That's a good idea.

The only downside that occurs to me would be the disappearance of `warn`. That is, today you get a helpful warning if you (say) accidentally type `Author:` instead of `Authors:`.

---

p.s. I don't understand this part:

> Right now, there is no good way to do that when using markdown, meaning I'd have to write all of my posts with scribble.

What would be the good way, how would you do it in Scribble?

LeifAndersen:
> The only downside that occurs to me would be the disappearance of warn. 

True. Although perhaps that could be configured in the blog's frog.rkt config file? Perhaps something like `enabled meta-tags`?

> What would be the good way, how would you do it in Scribble?

Okay, so probably not the best way in the world. And I'd have to double check how frog renders scribble files, but what I was thinking was having the template post.html file use a parameter defined in a different racket file, and having the blog post setting that parameter.

sorawee:
[...]

PR #242, opened 2019-01-09
Arbitrary metadata, paving the way, Pt. I

toothbrush:
The context for this PR is the discussion in #214.  To be clear, i don't really expect this PR to be accepted, but i would like to use it as a way to gather input on what i'm trying to do.  Also, apologies for my crude hacking, my Racket is rusty and probably needs restringing.

I spent quite a while staring at the Frog code, in the hope that i could help myself.  I basically want a reliable way to set a variable inside my blog post (whether it be Markdown or Scribble) and refer to it in `page-template.html`.  I found that i could kinda sorta abuse tags like `Foo:3` and regex them out in the template (at least i could use unmodified Frog), but that was really not helping me sleep at night.  So i thought i'd try to allow either arbitrary fields in https://github.com/greghendershott/frog/blob/08fa3db7cf45ac648ddd194d9051268dd0601519/frog/private/posts.rkt#L137 (but that would mean probably modifying `meta-data`'s return type.. everywhere) or alternatively take the path of least resistance and make the Racket list with tags available in the template by adding it to the dictionary here https://github.com/greghendershott/frog/blob/08fa3db7cf45ac648ddd194d9051268dd0601519/frog/private/posts.rkt#L261 which is fed into the page's environment with some macro magic here https://github.com/greghendershott/frog/blob/2bc070a60333e1316ff713f9b9b577a7c27e54b3/frog/private/template.rkt#L58  Forgive me, i know you know this, but i'm providing verbosity and links for posterity and those following along at home.

Anyway, modifying that hash turned out not to work, and after struggling on a few separate occasions and finally manually expanding the macro magic which `let`s the hash into the page's namespace, i saw that list values were getting turned from `'(happy list)` into `(happy list)`, which of course made Racket sad in the context of:

```Racket
(eval
 '(let ((boo ("flux" "flad")))
    ...)))
```

My commit basically puts back the magical `quote` to turn the value of `boo` back into a list.  Is there a better way to do this?  As best as i can tell i haven't broken anything by doing this.

### DIFF
--- a/frog/private/posts.rkt
+++ b/frog/private/posts.rkt
@@ -266,6 +266,7 @@
                'date-struct (date->date-struct date)
                'date        (~> date date->xexpr xexpr->string)
                'tags        (~> tags tags->xexpr xexpr->string)
+               'tag-list    tags
                'authors     (~> tags author-tags->xexpr xexpr->string)
                'date+tags   (~> (date+tags->xexpr date tags) xexpr->string)
                'content     body


greghendershott:
So it's been some years since I thought much about this code, but: I think the assumption has been that metadata values are always strings. Strings are "self-quoting". So it wasn't necessary to do what commit 47377e6 does. The commit seems innocent enough. What I don't know is the other consequences if any of allowing non-string values for metadata, generally??

Issue #209, opened 2018-03-01
Default syntax-highlight #:python-executable on Windows


greghendershott:
Arising from #208.

[`syntax-highlight`'s `#:python-executable`](https://github.com/greghendershott/frog/blob/master/frog/enhance-body.rkt#L25) defaults to `"python"`, always. Instead, the default should vary and be `"python.exe"` when on Windows.

Gist of it would be something like `(if (eq? (system-type) 'windows) "python.exe" "python")`.

> **Note**: This would be an easy item for a first-time contributor, and/or someone using Frog on Windows.

PR #251, opened 2019-06-22
use python.exe as default python executable name in windows

jeff66ruan:
Hi, 

This is for the [issue 209](https://github.com/greghendershott/frog/issues/209)


greghendershott:
Thanks!

PR #256, opened 2020-05-24
Default python executable should be python3

wlach:
On Mac (still) the default python is python2, but it's python3 that you almost
certainly want. Make this the default.

### DIFF
--- a/example/frog.rkt
+++ b/example/frog.rkt
@@ -15,8 +15,8 @@
   ;; Here we pass the xexprs through a series of functions.
   (~> xs
       (syntax-highlight #:python-executable (if (eq? (system-type) 'windows)
-                                                "python.exe"
-                                                "python")
+                                                "python3.exe"
+                                                "python3")
                         #:line-numbers? #t
                         #:css-class "source")
       (auto-embed-tweets #:parents? #t)


bremner:
This would be desirable on Debian and derivatives (e.g. Ubunto as well). The change for windows looks a bit suspect, but I don't know, I'm not a Windows user.

greghendershott:
Preface: I no longer use Frog myself. It's in low-maintenance mode. I have limited time to delve into this, including testing with various permutations of operating systems and Python versions.

`example/frog.rkt` is just that -- an example. Some users won't use Pygments at all. Those who do install and use it, will know more about the version of Pygments they installed and what version(s) of Python it needs.

So I'm not sure there's any one correct default here, and again this is just an example configuration?

bremner:
Greg Hendershott <notifications@github.com> writes:

> Preface: I no longer use Frog myself. It's in low-maintenance mode. I have limited time to delve into this, including testing with various permutations of operating systems and Python versions.
>
> `example/frog.rkt` is just that -- an example. Some users won't use Pygments at all. Those who do install and use it, will know more about the version of Pygments they installed and what version(s) of Python it needs.
>
> So I'm not sure there's any one correct default here, and again this is just an example configuration?

Kindof offtopic, but since you mention example configuration...

What I would really like is a clean way to override the example
configuration that gets installed with raco frog --init. I'm using frog
in a class where it is preinstalled in a VM image. Currently my only
idea is edit the files in the installed package, which seems kindof
gross. I would be happy to work on my own feature request, but I'm a bit
at sea as what to do.

d

wlach:
> Preface: I no longer use Frog myself. It's in low-maintenance mode. I have limited time to delve into this, including testing with various permutations of operating systems and Python versions.
> 
> `example/frog.rkt` is just that -- an example. Some users won't use Pygments at all. Those who do install and use it, will know more about the version of Pygments they installed and what version(s) of Python it needs.
> 
> So I'm not sure there's any one correct default here, and again this is just an example configuration?

FWIW this was a bit of a gotcha and it took me some time to figure out what was going on, which is what motivated me to file the PR. It's possible that landing this would just cause another set of issues. I too don't really have time or inclination to explore this further, maybe the best thing (if it's ok with the maintainer) is just to leave this open for now as a warning to others. At some point in the hopefully-not-too-distant-future, python2 will become rare.

lojic (commented):
I recommend this change. Having `python` instead of `python3` caused me some trouble initially on Mac OS. I'm not familiar with the situation on Windows, so if that's questionable, the pull request could be modified to only change non-Windows.

dertuxmalwieder:
On Windows, it would be “py -3”, preferably.

PR #259, opened 2020-08-19
use a color that's guaranteed to work for all scribble renderers

### LINKED ISSUE #258: PDF documentation build fails
Running `raco setup --doc-pdf x frog` generates pdf documentation in the `x` directory, except that the latex code fails. I think this is the relevant output that shows the error:

```
LaTeX Warning: Hyper reference `t:x28def_x28x28lib_x22frogx2fparamsx2erktx22x29
_currentx2dpermalinkx29x29' on page 25 undefined on input line 2070.


! LaTeX Error: Undefined color `grey'.

See the LaTeX manual or LaTeX Companion for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.2084 ... shell for literal \Scribtexttt{$\sim$}}
                                                  \end{SingleColumn}\end{SVe...

You're in trouble here.  Try typing  <return>  to proceed.
If that doesn't work, type  X <return>  to quit.

```

rfindler:
closes #258

### DIFF
--- a/frog/frog.scrbl
+++ b/frog/frog.scrbl
@@ -16,7 +16,7 @@
           scribble/extract)
 
 @(define (grey . contents)
-   (elem #:style (style #f (list (color-property "grey")))
+   (elem #:style (style #f (list (color-property (list 127 127 127))))
          contents))
 
 @(define (comment . contents)


rfindler:
Thanks!

PR #262, opened 2025-09-29
update metadata extracton for newer Scribble

mflatt:
Never versions of Scribble put document content in a `section` element.

I'm not sure that adjusting `meta-data` is the right approach. It could be that `adjust-scribble-html` in `read-scribble.rkt` is the right thing to change, instead.

### DIFF
--- a/frog/private/posts.rkt
+++ b/frog/private/posts.rkt
@@ -120,7 +120,8 @@
     [`(,(or `(pre () (code () . ,metas)) ;Markdown
             `(pre () . ,metas)           ;Markdown
             `(pre . ,metas)              ;Markdown
-            `(p () . ,metas))            ;Scribble
+            `(p () . ,metas)             ;Scribble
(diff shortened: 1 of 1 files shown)


mflatt:
My original PR avoided an error, but it seems that I didn't look at the generated output well enough. Documents in a `section` element turn into empty documents.

A change that worked for me was
```
--- a/frog/private/read-scribble.rkt
+++ b/frog/private/read-scribble.rkt
@@ -74,7 +74,12 @@
      '()]))
 
 (define (adjust-scribble-html xs img-uri)
-  (for/list ([x (in-list xs)])
+  (define xs/no-section
+    (match xs
+      [`((section ,_ . ,xs))
+       xs]
+      [_ xs]))
+  (for/list ([x (in-list xs/no-section)])
     (xexpr-map
      (lambda (x _)
        (list
```

Should I submit a new PR, or would a different process work better?

Thanks!

